PR #19, opened 2022-10-29
[PR1] DEV-9 integration up to console logging

jmaralo:
ChangoGUI MVP 1 that implements everything up to console logging. The next iteration will log all received data to a log file.

I've made significant changes to the file structure with the intention of making it easier to modify by decoupling some of the previous code we had. A detailed explanation follows:

The previous iterations had a **core problem** with their design: **direct communication between domain entities**, that is, different **components knew too much about each** other (making common import cycle errors). Most components **lacked a proper application layer** that would hide domains from each other and instead we opted to impor directly structs from other domains (e.g. we imported the PacketParser PacketUpdate from DataTranfer domain, which is wrong in two levels: first, this is something that should be performed on the application layer (relationship between different domains should ocurr on the application layer), not on the domain, and second, the DataTransfer domain was too coupled with the PacketParser domain).

Thus, I think we need to rething how we structure our code and how we couple it _carefully_, and this is my proposal. What i did to solve this problem focus mainly in two things: _use of interfaces_ and _changes to file structure_.

The first one is pretty straight forward: instead of importing structs from different domains, we define **common interfaces on the application layer** that then other components can import and use. For example, one of the changes I propose is relative to the ExcelParser: previously it defined multiple DTOs that we imported on multiple parts of the code in order to use them, now instead of importing those structs directly (which are now defined on the domain layer) we define a set of interfaces (that have functions such as Name(), ID(), etc.) that we import wherever we need to use them.

Then, the second problem is a matter of really thinking what each thing should be. The infrastructure layer should communicate all our layers with the outside world, the domain should be independent components that only implement logic relative to those components and the application layer should couple the multiple domain entities. Continuing the above example, the previous ExcelParser DTOs are now really part of the domain layer since their job has been taken by the common interfaces and the DTOs are now an internal data structure to the ExcelParser.

[This diagram](https://www.figma.com/file/wHoqjXHGmmMVowv7XHn4Nu/DEV-9-comparison?node-id=0%3A3) ilustrates the differences before and after.

Also it's reflected in the code: not a single component domain depends on other components domain, only the application layer coordinates component interaction, abstacting useless details and decoupling most of the code.

_Sorry for the long PR_, if there is something you dont understand or dissagre with please tell me, this is just my take on how we could make the backend better to mantain and iterate.

**TLDR: The previous integration made boundaries between components too fuzzy and abstract, making further improvement and debugging harder. The approach I propose creates clear boundaries between components by using interfaces and a strong layer model**

P.S.: This PR also fixes a bug with the ExcelRetriever, which would panic if it found  tables with a different amount of rows


jmaralo:
All code is the same except for some bug fixes that I made to some parts (mainly ExcelRetriever). Also I changed some names, so make sure to tell me if you would prefer another name for some structs or files or anything

smorsua (commented):
Using an interface to connect different domains is a great idea, but in our case I don't think it makes much difference. I say this because the interfaces you defined are basically getters for the properties in the structs each domain uses. This kind of implementation would make sense if the modules where to be used independently in multiple programs.

Nonetheless, I agree that it's not clear what structs should be used in different modules. From my understanding, the domain is the part of the module that is exposed for others to use, that's why i don't think its wrong for both the infrastructure of said module and other modules to import the same domain. I think the ExcelAdapter should be restructured as it doesn't follow the domain infra structure at all, and then the relationships between ExcelAdapter and the rest

jmaralo:
The thing is that we don't only have two layers, we have three: infrastructure, domain and application. The infrastructure communicates with the outside world (the pod, the google sheets API, etc.), the domain is a layer where we define each module internal logic and how it works (independently from the rest) and the application layer is where different modules interact with each other and the domain entities are instantiated and used.

What you say about interfaces is right, most of them are only getters, but you also state:

> This kind of implementation would make sense if the modules where to be used independently in multiple programs

That is the core idea of using interfaces and spliting code into layers: so that each part can be interchanged with something else without having to change multiple parts of the code to accommodate for that new feature, fix or whatever. **Each component is independent** and we can mix and match.

smorsua:
I would use interfaces only in the functions that connect different modules (return value on one module and parameter in another). But within a modules, I wouldn't use interfaces in the domain struct, so per example the Board struct in ExcelParser wouldn't have interfaces as properties, just the children structs.

I'm worried this will lead to a lot of code duplicity, having to write an interface for every struct in the domain. I rather have the structs be the contract between different modules.

jmaralo:
we deemed that the Interfaces weren't useful, so I removed them. The only thing left to talk about is the file structure and what each layer should really do

lipezaballa:
I would like to meet and see these changes in person to find out more about them, and I would also like to see what problem ExcelRetriever gave

smorsua (approved):

lipezaballa (approved):

PR #20, opened 2022-10-31
[PR1] DEV-11 Packet parser returns hex value and changes in podData structs

smorsua:
We are going to show the hexadecimal value of the packets in the frontend, so I made the necessary changes to be able to send that information.

I also separated the timestamp from the Packet as timestamp is not really a Packet property.


jmaralo (approved):

jmaralo:
Change the PR title to [PR1] DEV-11 ...

jmaralo (approved):

lipezaballa (approved):

PR #22, opened 2022-10-31
[PR1] DEV-18 Added channel to DataTranfer

smorsua:
Small change so DataTransfer outputs packages through a channel. Both the logger and the DataWebSocket will use the packets outputed by this channel.

### DIFF
--- a/src/DataTransfer/application/data_transfer.go
+++ b/src/DataTransfer/application/data_transfer.go
@@ -1,32 +1,30 @@
 package application
 
 import (
-	"time"
-
 	"github.com/HyperloopUPV-H8/Backend-H8/DataTransfer/domain"
 	excelParser "github.com/HyperloopUPV-H8/Backend-H8/Shared/ExcelParser/domain/board"
 	packetParser "github.com/HyperloopUPV-H8/Backend-H8/Shared/PacketAdapter/domain"
-
-	"github.com/davecgh/go-spew/spew"
 )
 
 type DataTransfer struct {
-	data domain.PodData
+	data          domain.PodData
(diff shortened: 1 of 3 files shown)


lipezaballa (approved):

src/DataTransfer/application/data_transfer.go
```diff
@@ -1,32 +1,30 @@
...
-
-		spew.Dump(packet)
-		<-time.After(time.Second)
-	}
+	go func() {
+		for {
+			update := getPacketUpdate()
+			dataTransfer.data.UpdatePacket(update)
+			packetTimestampPair := dataTransfer.data.GetPacket(update.ID)
+			dataTransfer.PacketChannel <- packetTimestampPair.Packet
```
jmaralo, line 27: this will end up lagging behind and sending packets that are not up to date because this will recieve at a much faster rate than the websocket, which will process 100 packets a second as of now. If we aim for a raw reading speed of 30mb/s and the average packet is 24 bytes (to say something, i don't know the exact size) this means we recieve over a thousand packets each second at most, but the websocket will consume them at a rate of 100 each second, so each second we are lagging 900 packets behind (this is some quick math that might be incorrect but you get the idea)

jmaralo (commented):
my only concern is what I say to you in the comment

src/DataTransfer/application/data_transfer.go
jmaralo: ok, i've been thinking it a bit more and this makes sense, we'll just need to create a dump somewhere so we dont clog the channel

PR #23, opened 2022-10-31
[PR1] DEV-17 Added OrderTransfer

smorsua:
Added OrderTransfer, which sends orders receive in a WS to the boards. The OrderWebAdapter has to be moved to the server/web_adapters folder once its available.

### DIFF
--- a/src/OrderTransfer/domain/order.go
+++ b/src/OrderTransfer/domain/order.go
@@ -1,6 +1,0 @@
-package domain
-
-type Order struct {
-	Id     uint16
-	Fields map[string]float64
-}
--- a/src/OrderTransfer/domain/order.go
+++ b/src/OrderTransfer/domain/order.go
@@ -1,0 +1,6 @@
+package domain
+
+type Order struct {
+	Id     uint16
+	Fields map[string]float64
+}
--- a/src/OrderTransfer/order_transfer.go
+++ b/src/OrderTransfer/order_transfer.go
@@ -1,0 +1,55 @@
+package ordertransfer
+
+import (
+	"log"
+	"strconv"
+
+	"github.com/HyperloopUPV-H8/Backend-H8/OrderTransfer/domain"
+)
+
+type OrderTransfer struct {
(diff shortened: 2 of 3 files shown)


src/OrderTransfer/order_transfer.go
```diff
@@ -0,0 +1,55 @@
...
+		for orderWebAdapter := range orderTransfer.OrderWebAdapterChannel {
+			order := getOrder(orderWebAdapter)
+			sendOrder(order)
+		}
+	}()
+}
+
+func getOrder(orderWA OrderWebAdapter) domain.Order {
+	id, err := strconv.Atoi(orderWA.Id)
+
```
lipezaballa, line 31: Why do you need to pass to int, can't you pass directly to uint16?

lipezaballa (approved):

src/OrderTransfer/order_transfer.go
smorsua: I can't do it directly. Atoi returns an int and another function called ParseInt returns a int64. In both cases I have to cast to uint16

src/OrderTransfer/domain/order.go
```diff
@@ -0,0 +1,6 @@
+package domain
+
+type Order struct {
+	Id     uint16
+	Fields map[string]float64
```
jmaralo, line 5: are we sure all values we get from an order ar float64? couldn't they be bool or enum?

src/OrderTransfer/order_transfer.go
```diff
@@ -0,0 +1,55 @@
+package ordertransfer
```
jmaralo, line 1: place OrderTransfer and OrderWebAdapter in their own application folder so that it follows the scheme we have for the rest of the modules

src/OrderTransfer/domain/order.go
smorsua: You're right, I'll make another commit

jmaralo (approved):

lipezaballa (approved):

PR #24, opened 2022-11-01
[PR1] DEV-14 added logger

jmaralo:
This PR adds the logger to the code base. I think it's done, but pls let me know if there is something missing


smorsua (approved):
Buen trabajo

PR #26, opened 2022-11-02
[PR1] DEV-13 Added streaming routines

smorsua:
This functions handle reading / writing to the websocket connections. The data is sent / received from the exterior using channels.


src/DataTransfer/domain/measurement/value/bool.go
```diff
@@ -6,7 +6,23 @@ import (
...
 type Bool bool
 
-func (b Bool) ToDisplayString() string {
+func (b Bool) ToDisplayUnitsString() string {
+	return b.toString()
+}
+
+func (b *Bool) GetPodUnits() string {
+	return ""
+}
```
lipezaballa, line 15: Is it correct that if this type of data does not need this method, implement it in blank? Is it not better to start all by default in blank and depending on the type use only the necessary ones?

src/Shared/StreamingHandlers/streaming.go
```diff
@@ -0,0 +1,42 @@
...
+			timeout := time.After(time.Millisecond * 20)
+		loop:
+			for {
+				select {
+				case packet := <-packetChannel:
+					adapter := newPacketWebAdapter(packet)
+					packetWebAdapterBuf = append(packetWebAdapterBuf, adapter)
+					if len(packetWebAdapterBuf) == 100 {
+						ws.WriteJSON(packetWebAdapterBuf)
+						break loop
```
lipezaballa, line 23: Run it only for the last packet, either here or with the timeout? Why?

src/Shared/StreamingHandlers/test.go
```diff
@@ -0,0 +1,62 @@
+package streaming
```
jmaralo, line 1: remove commented code

src/DataTransfer/domain/measurement/value/number/Number.go
```diff
@@ -14,7 +14,20 @@ func NewNumber(podUnitString string, displayUnitString string) *Number {
 	return &Number{value: 0, podUnits: podUnits, displayUnits: displayUnits}
 }
 
-func (i Number) ToDisplayString() string {
+func (n *Number) GetPodUnits() string {
```
jmaralo, line 17: why does this function receive a pointer?

jmaralo (changes requested):
just a few things here and there

src/DataTransfer/domain/measurement/value/bool.go
smorsua: Necesita implementarlo porque necesita implementar la interface Value. A cada tipo hay que implementar la función por separado, no puedes implementárselo a varios de golpe

jmaralo (dismissed):

lipezaballa (approved):

jmaralo (approved):

lipezaballa (approved):

PR #31, opened 2022-11-05
[PR1] DEV-22 add BOARD sheet identifier and fix bugs

lipezaballa:
filter excel "BOARD" sheets, use relatives path to credentials, corrected .env and json location to src. 
Corrected excelRetriever and tests problems, there were several problems, mostly in finding the height and filling the rows.
I have noted two minor problems that excel has, and there are several commented lines that explain a bit what the methods return because by changing Juan and implementing it with columns, we have a matrix transposed to what we are used to, and it helps when reading the code later on.


jmaralo:
I just added a commit to fix some bugs with the logic of excel reader.

What happend was that when it found a table that was on "the edge" (the bottom right corner) it tried to parse it iterating the elements of two axes until it found blank space, then it would update the table bound with those values. Because once we reach that corner there is no more data to be read the loop ended without finding empty cells because for the excel there were no more cells in the sheet and it would say that the table had 0 rows and 0 cols.

Also we need to change the logic because right now it only checks the first column and the first row to get the dimensions, but maybe some elements are missing or columns have different sizes and we end up with an smaller table than real one, missing necessary information but thinking that what we got is all that exists on that table.

jmaralo:
This last commit should fix all known bugs

src/Shared/excel_retriever/excelRetriever.go
```diff
@@ -10,8 +10,8 @@ import (
 	"github.com/xuri/excelize/v2"
 )
 
-func GetExcel(fileName string, filePath string) domain.Document {
-	infra.FetchExcel(os.Getenv("SPREADSHEET_ID"), fileName, filePath)
+func GetExcel(fileName string, filePath string, credentialsPath string) domain.Document {
+	infra.FetchExcel(os.Getenv("SPREADSHEET_ID"), fileName, filePath, credentialsPath)
```
smorsua, line 14: Make the SPREADSHEET_ID a parameter like the credentialsPath

src/Shared/excel_retriever/infra/downloadExcel.go
```diff
@@ -14,15 +14,15 @@ import (
...
+	fileBuf := downloadExcel(spreadsheetID, fileName, credentialsPath)
 	saveFileToPath(fileBuf, fileName, filePath)
 }
 
-func downloadExcel(spreadsheetID string, fileName string) []byte {
+func downloadExcel(spreadsheetID string, fileName string, credentialsPath string) []byte {
 	ctx := context.Background()
-
-	driveService, err := drive.NewService(ctx, option.WithCredentialsFile(os.Getenv("SECRET_FILE_PATH")))
+	filePath := credentialsPath + os.Getenv("SECRET_FILE_PATH")
```
smorsua, line 24: Receive SECRET_FILE_PATH as parameter

src/Shared/excel_retriever/infra/readExcel.go
```diff
@@ -1,93 +1,126 @@
...
+	document := domain.Document{
+		Sheets: sheets,
+	}
+	return document
+}
+
+func ParseSheets(file *excelize.File) map[string]domain.Sheet {
+	sheets := make(map[string]domain.Sheet)
+	boards := sheetsFilter(file.GetSheetMap())
+	fmt.Print("boards filtered: ")
```
smorsua, line 26: Remove prints

src/Shared/excel_retriever/infra/readExcel.go
```diff
@@ -1,93 +1,126 @@
...
+	fmt.Println(boards)
+	for _, name := range boards {
+		cols, err := file.GetCols(name)
+		if err != nil {
+			log.Fatalf("get rows: %s\n", err)
```
smorsua, line 31: Make the error say "error gettings columns"

src/Shared/excel_retriever/infra/readExcel.go
```diff
@@ -1,93 +1,126 @@
...
+	return document
+}
+
+func ParseSheets(file *excelize.File) map[string]domain.Sheet {
+	sheets := make(map[string]domain.Sheet)
+	boards := sheetsFilter(file.GetSheetMap())
+	fmt.Print("boards filtered: ")
+	fmt.Println(boards)
+	for _, name := range boards {
+		cols, err := file.GetCols(name)
```
smorsua, line 29: Make a function called getCols and return only the columns. The error handling of file.GetCols would be done inside this new function

src/Shared/excel_retriever/infra/readExcel.go
```diff
@@ -1,93 +1,126 @@
...
+			if strings.HasPrefix(cell, tablePrefix) {
+				end := findTableEnd(cols, i, j)
+				tables[strings.TrimPrefix(cell, tablePrefix)] = [4]int{i, j, i + end[0], j + end[1]}
+			}
+		}
+	}
+	return tables
+}
+
+func findTableEnd(cols [][]string, firstCol int, firstRow int) (bound [2]int) {
```
smorsua, line 74: Split into two functions: findTableWidth y findTableHeight

src/Shared/excel_retriever/infra/readExcel.go
```diff
@@ -1,93 +1,126 @@
...
+			maxHeight = bound[1]
+		}
+	}
+	bound[1] = maxHeight
+
+	return
+}
+
+func parseTable(name string, cols [][]string, bound [4]int) domain.Table {
+	fmt.Print("Parse Table, bound[3]-bound[1]-2: ")
```
smorsua, line 107: remove prints

smorsua (changes requested):

src/Shared/excel_retriever/excelRetriever.go
```diff
@@ -2,16 +2,16 @@ package excelRetriever
...
 
 	"github.com/HyperloopUPV-H8/Backend-H8/Shared/excel_retriever/domain"
 	"github.com/HyperloopUPV-H8/Backend-H8/Shared/excel_retriever/infra"
 	"github.com/xuri/excelize/v2"
 )
 
-func GetExcel(fileName string, filePath string) domain.Document {
-	infra.FetchExcel(os.Getenv("SPREADSHEET_ID"), fileName, filePath)
+func GetExcel(spreadSheetID string, fileName string, filePath string, credentialsPath string) domain.Document {
+	//os.Getenv("SPREADSHEET_ID")
```
jmaralo, line 13: commented code

src/Shared/excel_retriever/infra/downloadExcel.go
```diff
@@ -14,15 +14,15 @@ import (
...
+	fileBuf := downloadExcel(spreadsheetID, fileName, credentialsPath)
 	saveFileToPath(fileBuf, fileName, filePath)
 }
 
-func downloadExcel(spreadsheetID string, fileName string) []byte {
+func downloadExcel(spreadsheetID string, fileName string, credentialsPath string) []byte {
 	ctx := context.Background()
-
-	driveService, err := drive.NewService(ctx, option.WithCredentialsFile(os.Getenv("SECRET_FILE_PATH")))
+	//filePath := credentialsPath + os.Getenv("SECRET_FILE_PATH")
```
jmaralo, line 24: commented code

src/cmd/MVP-1/main.go
```diff
@@ -1,8 +1,8 @@
 package main
```
jmaralo, line 1: we shouldn't be editing the MVP-1 main when we changes change the code, the MVP-1 was a previous varsion and all of this is for the next MVP

smorsua (dismissed):

jmaralo (changes requested):
just a few minor things before merging

jmaralo (approved):

smorsua (approved):

PR #35, opened 2022-11-06
[PR1] DEV-30 MVP-2 Backend

jmaralo:
This PR creates the main file that orchestrates the backend for the MVP-2 and fixes some conflicts between modules


jmaralo:
this PR is a certified programmer moment :)

jmaralo:
Will separate the front from the back in future commits, as of now the working frontend will stay in cmd/MVP-2/static so that it's easier to distribute

src/Shared/Logger/infra/file_handler.go
```diff
@@ -0,0 +1,73 @@
...
+	"path"
+)
+
+type FileData struct {
+	handle *os.File
+	values []ValueTimestamp
+}
+
+func NewFile(parentPath string, name string) *FileData {
+	file, err := os.Create(path.Join(parentPath, fmt.Sprintf("%s.csv", name)))
```
lipezaballa, line 16: Why do we use .csv?

smorsua (approved):

src/Shared/Logger/infra/logger.go
```diff
@@ -1,137 +1,62 @@
...
+	}()
+}
+
+func (log Logger) record() {
+	log.currDir = NewDir(path.Join(log.baseDir, time.Now().Format("2006-01-02_15-04-05")))
+loop:
+	for {
+		select {
+		case <-log.ticker.C:
+			log.currDir.Dump()
```
lipezaballa, line 45: what does "lod.currDir.Dump" do?

src/Shared/StreamingHandlers/streaming.go
```diff
@@ -1,67 +1,63 @@
...
+			}
+		}
+	}()
+}
+
+func OrderSocketHandler(ws websocket.Conn, orderWAChannel chan OrderWebAdapter) {
+	go func() {
+		for {
+			orderWA := OrderWebAdapter{}
+			ws.ReadJSON(orderWA)
```
lipezaballa, line 35: what this function does, where it waits for an order to arrive to send?

lipezaballa (approved):

src/Shared/Logger/infra/file_handler.go
smorsua: We spoke to electromagnetics and agreed on that format. They can import it in matlab easily.

src/Shared/StreamingHandlers/streaming.go
smorsua: ReadJSON waits for data to come from the websocket (ws) and when it gets it, it puts it inside orderWA

src/Shared/Logger/infra/logger.go
smorsua: Each file has a corresponding array that fills up with entries, and every once in a while the content from these arrays is copied into the corresponding file. Dump puts the contents of each array in the corresponding file.

PR #37, opened 2022-11-06
[PR1] DEV-33 Fixed Count and CycleTime problem, fixed MVP-1

smorsua:
The issue with the UpdatePacket was that the CycleTime and the Count weren't being updated. This happend because when you do a range loop, each item in the slice/map is copied by value, not reference. This meant that the original items in PodData weren't being affected by the changes inside the range loop. To fix this, I made the items in PodData (boards, packets) pointers.

All changes to the excel_retriever files should be overlooked as they are going to be reviewed in the integrations/DEV-30 PR.


jmaralo (approved):
probably will cause merge conflicts with DEV-30, but as it is looks great

PR #40, opened 2022-11-07
[PR2] DEV-30 Refactored main (MVP-2)

smorsua:
This PR restructures the main from MVP-2 so the packets are received from a channel from data_transfer. I also changed the http server so the types are more accurate.

Most files that appeared changed are the result of changing a couple of folder names and imports, so I'll put here the files I actually changed:

- main.go (MVP-2)
- data_transfer.go
- logger.go
- http.go
- websocket-go
- message.go
- message_transfer.go
- main.go (MVP-1)


smorsua:
The comment on streaming.go is a code that we will need in the future.

src/Shared/Server/infra/http.go
```diff
@@ -16,18 +16,18 @@ const (
 type HTTPServer[D, O, M any] struct {
 	router      *mux.Router
 	page        spaHandler
-	PacketRecv  chan D
-	OrderSend   chan O
-	MessageRecv chan M
+	PacketChan  chan D
```
jmaralo, line 19: isnt this redundant? you already know this is a channel from the type

src/Shared/streaming_handlers/streaming.go
```diff
@@ -0,0 +1,63 @@
...
+// 						break loop
+// 					}
+// 				case <-timeout:
+// 					ws.WriteJSON(messageWebAdapterBuf)
+// 					break loop
+// 				}
+// 			}
+// 		}
+// 	}()
+// }
```
jmaralo, line 63: commented code

src/cmd/MVP-2/main.go
```diff
@@ -29,33 +32,60 @@ func main() {
...
+	// 	for {
+	// 		payload := dataTransfer.Parse(packetAdapter.ReceiveData)
+	// 		select {
+	// 		case logger.EntryChan <- *payload:
+	// 		default:
+	// 		}
+	// 		dataTransfer.PacketTimestampChan <- payload.Packet
+	// 	}
+	// }()
+
```
jmaralo, line 75: commented code

jmaralo (changes requested):
just two things to make the code cleaner

src/Shared/Server/infra/http.go
smorsua: Yes but I could find a better name

jmaralo (approved):
lgtb

PR #44, opened 2022-11-09
[PR1] DEV-40 added Units module

jmaralo:
This PR adds the Units module to decouple the unit convertion system from the DataTransfer module and allow us to use it in multiple places.


src/Shared/units/infra/units.go
```diff
@@ -0,0 +1,53 @@
+package infra
+
+import (
+	"github.com/HyperloopUPV-H8/Backend-H8/Shared/units/domain"
+)
+
+type Units struct {
+	pod     map[string]domain.Operations
```
smorsua, line 8: Search for an alternative to make a pair. It's not a bad idea but its confusing

smorsua (changes requested):

smorsua (approved):

lipezaballa (approved):

PR #46, opened 2022-11-10
[PR1] DEV-42 add MessageTransfer

jmaralo:
This PR adds the MessageTransfer module.

Before merging I'd like to update PacketAdapter, but to prevent conflicts I first need the changes over on [DEV-41](https://github.com/HyperloopUPV-H8/Backend-H8/pull/45), so please check that one first


Losina24 (dismissed):
Pinta bien

src/message_transfer/domain/message.go
```diff
@@ -1,6 +1,23 @@
 package domain
 
+type id = uint16
+
 type Message struct {
-	Id     uint16
-	Fields map[string]string
+	id     id
+	values map[string]any
```
smorsua, line 7: Why do you used any? Messages only contain strings

src/message_transfer/presentation/websocket.go
```diff
@@ -0,0 +1,15 @@
+package presentation
+
+import (
+	"github.com/HyperloopUPV-H8/Backend-H8/Shared/server/infra/interfaces"
+	"github.com/HyperloopUPV-H8/Backend-H8/message_transfer/domain"
+)
+
+func MessageRoutine(socket interfaces.WebSocket, messages <-chan domain.Message) {
```
smorsua, line 8: Is the handling of ws.onError done somewhere else?

src/message_transfer/domain/message.go
jmaralo: right now FW says they'll only send strings, but maybe in the future they might send numbers or booleans as well. By using any they can now send whatever they want in a message

smorsua: Let's stick to the specifications, we can change it in the future

src/message_transfer/presentation/websocket.go
smorsua: Okay

smorsua (approved):

lipezaballa (approved):

PR #50, opened 2022-11-14
[PR1] DEV-47 add Server tests

jmaralo:
This PR adds test for the Server module.

### DIFF
--- a/src/Shared/Server/infra/log.go
+++ b/src/Shared/Server/infra/log.go
@@ -7,7 +7,7 @@ import (
 )
 
 func (server HTTPServer[D, O, M]) HandleLog(route string, loggerEnable chan<- bool) {
-	server.router.Handle(route, LogHandle{logEnable: loggerEnable}).Methods("PUT")
+	server.router.Handle(route, LogHandle{logEnable: loggerEnable}).Methods(http.MethodPut)
 }
 
 type LogHandle struct {
@@ -20,5 +20,13 @@ func (handler LogHandle) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		log.Fatalln("http server: log handle:", err)
 	}
 
-	handler.logEnable <- (string(body) == "enable")
+	go func() {
+		if string(body) == "enable" {
+			handler.logEnable <- true
+		} else if string(body) == "disable" {
+			handler.logEnable <- false
+		}
+	}()
+
+	w.Write([]byte{})
 }
--- a/src/Shared/Server/infra/server_test.go
+++ b/src/Shared/Server/infra/server_test.go
@@ -1,0 +1,79 @@
+package infra
+
+import (
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"path"
+	"reflect"
+	"testing"
+)
+
+func TestPage(t *testing.T) {
+	defaultStaticPath = path.Join("test", "resources")
+	t.Cleanup(func() {
+		defaultStaticPath = path.Join("static", "build")
+	})
+
+	server := New[any, any, any]()
+	server.HandleSPA()
+	go server.ListenAndServe()
+
+	client := http.Client{}
+
+	type testCase struct {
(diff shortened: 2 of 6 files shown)


lipezaballa (approved):

src/Shared/Server/infra/test/resources/index.js
```diff
@@ -0,0 +1 @@
+alert("hello world!")
```
lipezaballa, line 1: this is enough to check it?

lipezaballa: may be a description of what works rather than hello world would be better.

lipezaballa (approved):

src/Shared/Server/infra/test/resources/index.js
jmaralo: This code doesn't even need to work, it's just there to check that the server serves the correct files. It could perfectly be a text file with random text

src/Shared/Server/infra/log.go
```diff
@@ -20,5 +20,13 @@ func (handler LogHandle) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		log.Fatalln("http server: log handle:", err)
 	}
 
-	handler.logEnable <- (string(body) == "enable")
+	go func() {
```
querricalapepis, line 23: This is more related to the logger, but I'll put it here. The way the logger is setup right now, multiple logging sessions can be started at the same. Either we add ids to "logging sessions" so that multiple logging sessions can be coordinated at the same time or we set it up so that logging is shared between users. I'd go for the first options.

src/Shared/Server/infra/server_test.go
```diff
@@ -0,0 +1,206 @@
...
+	defaultStaticPath = path.Join("test", "resources")
+	serverAddr = "127.0.0.1:4002"
+
+	server := New[any, any, any]()
+	server.HandleWebSocketData("/backend/data", func(interfaces.WebSocket, <-chan any) {})
+	server.HandleWebSocketMessage("/backend/message", func(interfaces.WebSocket, <-chan any) {})
+	server.HandleWebSocketOrder("/backend/order", func(interfaces.WebSocket, chan<- any) {})
+	go server.ListenAndServe()
+
+	_, resp := wsClient("ws://" + serverAddr + "/backend/data")
```
querricalapepis, line 141: put "make ws request and check status switching protocol" in a separate function

querricalapepis (changes requested):

src/Shared/Server/infra/log.go
smorsua, line 23: This is more related to the logger, but I'll put it here. The way the logger is setup right now, multiple logging sessions can be started at the same. Either we add ids to "logging sessions" so that multiple logging sessions can be coordinated at the same time or we set it up so that logging is shared between users. I'd go for the first options.

src/Shared/Server/infra/server_test.go
smorsua, line 141: Put "make ws request and check status switching protocol" in a separate function

smorsua (changes requested):

src/Shared/Server/infra/log.go
jmaralo: currently we can only start a global logging session, multiple sessions can't be running at the same time because we only instantiate one logger and we use only one channel to toggle it. All routines receive the same channel and any one connected to the backend can start/stop the logger.

FW stated that no more than one person will be logging at the same time, so I wouldn't change it until they request it, specially if we want to deliver a new MVP soon (we need at least this week to test everything and put it all together).

src/Shared/Server/infra/server_test.go
jmaralo: ok

smorsua (approved):

lipezaballa (approved):

PR #51, opened 2022-11-16
[PR1] DEV-51 add PodData to Server

jmaralo:
This PR adds the PodData struct to the Server that is sent to the front when connecting


smorsua (approved):

smorsua:
I approved, but the PodData creation has to be handled somewhere else

lipezaballa (approved):

PR #54, opened 2022-11-29
[PR1] DEV-53 add concurrency to packet adapter

jmaralo:
This PR adds a concurrency to the Packet Adapter.

It uses a ring buffer to store a series of channels that sends packets to one of n goroutines. Each time a packet arrives we send a packet through the current buffer channel and advance the buffer by one. This only applies to basic data, not to messages and orders, those have their own routine.

I also performed some benchmarks and test on multiple modules, here is the speed at which each module runs on my machine:

goos: windows
goarch: amd64
cpu: AMD Ryzen 7 3700U with Radeon Vega Mobile Gfx

*pps: packets per second

- PacketParser: 648,929 pps
- Units: 15,112,588 pps
- Sniffer: TBD (we need a faster switch, the max speed we reached was 40mb/s)


smorsua (commented):
Good job implementing concurrency, please keep your PRs smaller.

jmaralo:
Closing because this gets obsolete after the structure refactor

PR #55, opened 2022-12-09
[PR1] DEV-70 add TransportController

jmaralo:
This is the TransportController module after the structure changes we proposed a few weeks ago.

There are two main components: Pipes (tcp connections) and the Sniffer (capture network traffic)

## Pipes
Each pipe represents and individual TCP connection to a board. The connection is a pointer to a TCPConn, if the pointer is nil this means the board is disconnected.

The only way to access the pipes is through the PipeHandle, which creates each pipe and manages connections. Currently it's configured so each pipe will listen for connections and at the same time tries to connect to the board to support a wide range of applications.

Once a pipe is created it will automatically manage connections and disconnections, notifying other modules when this happens. Reading happens automatically when the pipe is open, sending all the received information through a channel

Writing to a pipe will return a bool that indicates if the operation was successful.

## Sniffer
The sniffer uses pcap to capture network traffic on the specified device. We can specify a [BPF](https://en.wikipedia.org/wiki/Berkeley_Packet_Filter) to prevent unwanted traffic from being captured.

Each time a valid message is received it's sent through a channel.

We can also read the handle stats to see the amount of packets received and lost.

## TransportController
This is the part that puts it all together, it provides bindings to the underlying components and manages them.

We can specify some configuration options that will be used during the creation of the component, but once the TransportController is initialized further changes won't affect it (config is passed by value)


src/transport_controller/internals/pipe.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (pipe *Pipe) write(payload []byte) (success bool) {
+	if _, err := pipe.conn.Write(payload); err != nil {
+		pipe.close()
+		return false
+	}
+	return true
+}
+
+func getPipe(laddr *net.TCPAddr, raddr *net.TCPAddr, config models.Config) *Pipe {
```
smorsua, line 69: Rename to createPipe

smorsua (approved):

src/transport_controller/internals/pipe.go
```diff
@@ -0,0 +1,80 @@
...
+	"net"
+
+	"github.com/HyperloopUPV-H8/Backend-H8/transport_controller/models"
+)
+
+type Pipe struct {
+	conn   *net.TCPConn
+	laddr  *net.TCPAddr
+	raddr  *net.TCPAddr
+	config models.Config
```
lipezaballa, line 13: what is laddr and raddr for?

lipezaballa (approved):

src/transport_controller/internals/pipe.go
jmaralo: so it knows the addresses (laddr: local address, raddr: remote addres (the board)) when connecting to the board

PR #56, opened 2022-12-09
[PR2] DEV-70 add PacketParser

jmaralo:
The PacketParser is in charge of encoding and decoding packets based on the information it's provided.

There are various encoders and decoders for each type of value. The code is almost the same as the previous PacketParser, but it adds a method (AddPacket) which is used by the ExcelAdapter to build the struct from the excel.


src/packet_parser/packet_parser.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/HyperloopUPV-H8/Backend-H8/packet_parser/models"
+)
+
+type PacketParser struct {
+	descriptors map[uint16]models.PacketDescriptor
+	enums       map[string]models.Enum
+}
+
+func (parser *PacketParser) AddPacket(board string, ip string, desc excelAdapterModels.Description, values []excelAdapterModels.Value) {
+	if parser.descriptors == nil {
```
smorsua, line 21: Possible error should be handled outside this function.

src/packet_parser/packet_parser.go
```diff
@@ -0,0 +1,104 @@
...
+		log.Fatalf("packet parser: AddPacket: %s\n", err)
+	}
+
+	valueDescriptors := make([]models.ValueDescriptor, len(values))
+	for i, value := range values {
+		kind := value.Type
+		if strings.HasPrefix(kind, "ENUM") {
+			kind = "enum"
+			if parser.enums == nil {
+				parser.enums = make(map[string]models.Enum)
```
smorsua, line 36: Same with this

smorsua (changes requested):
Some functions have responsability that doesnt belong to them. Looks good overall anyways

src/packet_parser/packet_parser.go
lipezaballa: and where should it be?

lipezaballa (approved):

smorsua (approved):

PR #59, opened 2022-12-09
[PR5] DEV-70 add ExcelAdapter

jmaralo:
The ExcelAdapter downloads the excel and parses its contents.

This is the same as the previous ExcelAdapter, the only change is the name of the sheets and the Compile method, which takes various objects that can be built from the boards and puts them together


src/excel_adapter/excel_adapter.go
```diff
@@ -0,0 +1,53 @@
...
+	"path/filepath"
+
+	"github.com/HyperloopUPV-H8/Backend-H8/excel_adapter/internals"
+	internalModels "github.com/HyperloopUPV-H8/Backend-H8/excel_adapter/internals/models"
+	"github.com/HyperloopUPV-H8/Backend-H8/excel_adapter/models"
+	"github.com/xuri/excelize/v2"
+)
+
+const GlobalSheet = "Info"
+const AddressesTable = "addresses"
```
smorsua, line 15: Put each address in the corresponding board sheet

smorsua (approved):
LGTM!

src/excel_adapter/internals/excel_reader.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return models.Sheet{
+		Tables: tables,
+	}
+}
+
+func findTables(cols [][]string) map[string][4]int {
+	tables := make(map[string][4]int)
+	for i, col := range cols {
+		for j, cell := range col {
```
lipezaballa, line 67: may be separate the second for in another function?

src/excel_adapter/internals/excel_reader.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return -1
+}
+
+func findTableHeight(cols [][]string, firstCol int, firstRow int, width int) int {
+	maxHeight := 0
+	height := 0
+	for _, col := range cols[firstCol : width+firstCol] {
+		for j, cell := range col[firstRow+2:] { //firstRox+2: first data row
+			if cell == "" {
```
lipezaballa, line 102: same, another function, maybe not necessary, but more manageable

lipezaballa (approved):

PR #60, opened 2022-12-09
[PR6] DEV-70 add DataTransfer (after structure change)

jmaralo:
DataTransfer takes all the incoming data decoded by the packetParser and sends it to the front.

The main difference relative to the last version is that now we support multiple webSocket connections open at the same time and all of them will recieve the packets.

Also now PodData is inside this module

### DIFF
--- a/src/data_transfer/data_transfer.go
+++ b/src/data_transfer/data_transfer.go
@@ -1,0 +1,41 @@
+package data_transfer
+
+import (
+	"time"
+
+	"github.com/HyperloopUPV-H8/Backend-H8/data_transfer/models"
+	"github.com/gorilla/websocket"
+)
+
+type DataTransfer struct {
+	routines []chan<- models.PacketUpdate
+	Rate     time.Duration
+}
+
+func (dataTransfer *DataTransfer) HandleConn(socket *websocket.Conn) {
+	updates := make(chan models.PacketUpdate)
+	dataTransfer.routines = append(dataTransfer.routines, updates)
+
+	go func(socket *websocket.Conn, updates <-chan models.PacketUpdate, rate time.Duration) {
+		defer socket.Close()
+		buf := make(map[uint16]models.PacketUpdate)
+		ticker := time.NewTicker(rate)
+	loop:
+		for {
+			select {
+			case payload := <-updates:
+				buf[payload.ID] = payload
+			case <-ticker.C:
+				if err := socket.WriteJSON(buf); err != nil {
+					break loop
+				}
+			}
+		}
+	}(socket, updates, dataTransfer.Rate)
+}
+
+func (dataTransfer *DataTransfer) Broadcast(update models.PacketUpdate) {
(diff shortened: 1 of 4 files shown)


src/data_transfer/data_transfer.go
```diff
@@ -0,0 +1,41 @@
...
+type DataTransfer struct {
+	routines []chan<- models.PacketUpdate
+	Rate     time.Duration
+}
+
+func (dataTransfer *DataTransfer) HandleConn(socket *websocket.Conn) {
+	updates := make(chan models.PacketUpdate)
+	dataTransfer.routines = append(dataTransfer.routines, updates)
+
+	go func(socket *websocket.Conn, updates <-chan models.PacketUpdate, rate time.Duration) {
```
smorsua, line 19: Define function outside

src/data_transfer/models/pod_data.go
```diff
@@ -0,0 +1,96 @@
...
+
+	excelAdapterModels "github.com/HyperloopUPV-H8/Backend-H8/excel_adapter/models"
+)
+
+type PodData struct {
+	Boards map[string]Board `json:"boards"`
+}
+
+func (podData *PodData) AddPacket(board string, ip string, desc excelAdapterModels.Description, values []excelAdapterModels.Value) {
+	if podData.Boards == nil {
```
smorsua, line 16: Define this properties in a podData contructor PodData::New()

src/data_transfer/models/pod_data.go
```diff
@@ -0,0 +1,96 @@
...
+		podData.Boards = make(map[string]Board)
+	}
+
+	id, err := strconv.ParseUint(desc.ID, 10, 16)
+	if err != nil {
+		log.Fatalf("data transfer: AddPacket: %s\n", err)
+	}
+
+	dataBoard, ok := podData.Boards[board]
```
smorsua, line 25: No boards should be defined at first, so no need to check

src/data_transfer/models/pod_data.go
```diff
@@ -0,0 +1,96 @@
...
+	"strings"
+
+	excelAdapterModels "github.com/HyperloopUPV-H8/Backend-H8/excel_adapter/models"
+)
+
+type PodData struct {
+	Boards map[string]Board `json:"boards"`
+}
+
+func (podData *PodData) AddPacket(board string, ip string, desc excelAdapterModels.Description, values []excelAdapterModels.Value) {
```
smorsua, line 15: Separate this function into multiple functions.

smorsua (approved):

lipezaballa (approved):

src/data_transfer/models/pod_data.go
jmaralo: Boards is defined, here I'm checking if the board of the packet exsist (which may not be the case) and adding a new entry or updating the existing one (we cannot init this beforehand because with this approach we don't have a board list in main)

PR #61, opened 2022-12-09
[PR7] DEV-70 add MessageTransfer

jmaralo:
MessageTransfer sends all the faults and warnings through a specific channel so the front can differentiate them from regular packets.

Most of the code is the same as before.


src/message_transfer/message_transfer.go
```diff
@@ -0,0 +1,47 @@
...
+			} else if msg, ok = update.Values["fault"]; ok {
+				message = models.Message{
+					ID:          update.ID,
+					Description: msg,
+					Type:        "warning",
+				}
+			}
+
+			if err := socket.WriteJSON(message); err != nil {
+				return
```
lipezaballa, line 37: nothing happens if there is an error?

lipezaballa (approved):

src/message_transfer/message_transfer.go
jmaralo: if there is an error we understand the connection is down and close the socket

src/message_transfer/message_transfer.go
```diff
@@ -0,0 +1,47 @@
...
+			}
+		}
+	}(socket, updates)
+}
+
+func (messageTransfer *MessageTransfer) Broadcast(update dataTransferModels.PacketUpdate) {
+	for _, routine := range messageTransfer.routines {
```
smorsua, line 44: Consider keeping an array of WS instead of multiple channels, it adds a middle layer thats not necessary.

smorsua (commented):
Check the comment

src/message_transfer/message_transfer.go
jmaralo: it's not the same, here the channels are made because each socket runs in its own goroutine and I need a way to send the messages to those goroutines

src/message_transfer/message_transfer.go
```diff
@@ -1,17 +1,19 @@
...
+	"time"
+
 	dataTransferModels "github.com/HyperloopUPV-H8/Backend-H8/data_transfer/models"
 	"github.com/HyperloopUPV-H8/Backend-H8/message_transfer/models"
 	"github.com/gorilla/websocket"
 )
 
 type MessageTransfer struct {
-	sockets map[string]*websocket.Conn
+	sockets map[int64]*websocket.Conn
```
smorsua, line 12: For the id, use this or similar: https://github.com/kjk/betterguid

PR #62, opened 2022-12-09
[PR8] DEV-70 add OrderTransfer

jmaralo:
OrderTransfer is in charge of receiving orders from the front.

The only thing that has changed is that now OrderData is inside OrderTransfer (previously it was inside HTTPServer)


src/order_transfer/order_transfer.go
```diff
@@ -0,0 +1,23 @@
+package order_transfer
+
+import (
+	"github.com/HyperloopUPV-H8/Backend-H8/order_transfer/models"
+	"github.com/gorilla/websocket"
+)
+
+type OrderTransfer struct {
+	Dump chan<- models.Order
```
smorsua, line 9: Change name

smorsua (commented):
Good overall

lipezaballa (approved):

smorsua (approved):

PR #64, opened 2022-12-09
[PR10] DEV-70 add LogHandle and go.mod

jmaralo:
LogHandle manages all the loging performed by the backend.

It stores the raw decoded data from the pod. It creates a different file for each measure and each file is a .csv where the first column is the timestamp (nanos from UTC) and the second one the actual value.

The logger is first runned (calling Run) and after that it can be enabled / disabled with the Start and Stop methods


src/log_handle/log.go
```diff
@@ -0,0 +1,95 @@
...
+			select {
+			case <-log.source:
+			default:
+			}
+			continue
+		}
+		select {
+		case value := <-log.source:
+			log.buffer[value.Name] = append(log.buffer[value.Name], value)
+			if len(log.buffer[value.Name]) > int(log.config.DumpSize/log.config.RowSize) {
```
smorsua, line 45: To calculate the in memory size of the buffer, look into binary.Size

src/log_handle/log.go
```diff
@@ -0,0 +1,95 @@
...
+	defer log.dumpMx.Unlock()
+	for value, buffer := range log.buffer {
+		log.writeCSV(value, buffer)
+	}
+	log.buffer = make(map[string][]models.Value)
+}
+
+func (log *LogHandle) writeCSV(value string, buffer []models.Value) {
+	file, ok := log.dumps[value]
+	if !ok {
```
smorsua, line 76: File should already be created once logging for the value starts.

smorsua (approved):
Good

smorsua (commented):
Good

src/log_handle/log.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/HyperloopUPV-H8/Backend-H8/log_handle/models"
+)
+
+type LogHandle struct {
+	source <-chan models.Value
+	buffer map[string][]models.Value
+	config models.Config
+	dumpMx sync.Mutex
+	dumps  map[string]*os.File
+}
```
lipezaballa, line 21: what is dumps for?

lipezaballa (approved):

smorsua (approved):

PR #67, opened 2022-12-13
[PR1] DEV-89 improve logger performance

jmaralo:
The previous logger implementation was innefficient and consumed a lot of resources, this new implementation is faster and doesn't block as much, allowing the back to reach its performance requirements.

Current tests suggest the back can handle all the traffic that the computer is able to send through the loopback interface, further testing with actuall boards and network interfaces will actually tell us if we meet the performance requirements or not.

### DIFF
--- a/src/log_handle/log.go
+++ b/src/log_handle/log.go
@@ -15,61 +15,71 @@ import (
 )
 
 type LogHandle struct {
-	bufferMx sync.Mutex
+	flushing bool
 	buffer   map[string][]models.Value
 	config   models.Config
-	filesMx  sync.Mutex
+	fileMx   sync.Mutex
 	files    map[string]*os.File
+	done     chan struct{}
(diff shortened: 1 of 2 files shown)


src/log_handle/log.go
```diff
@@ -15,61 +15,71 @@ import (
...
-		})
-		if len(logger.buffer[name]) > int(logger.config.DumpSize/logger.config.RowSize) {
-			dump = true
+func (logger *LogHandle) run() {
+	logger.running = true
+	defer func() { logger.running = false }()
+	for {
+		select {
+		case update := <-logger.config.Updates:
+			for name, value := range update {
```
smorsua, line 44: Make each for a function

smorsua (approved):

lipezaballa (approved):

PR #69, opened 2023-01-03
[PR1] DEV-101 Backend V1.0

jmaralo:
This is the first release of the backend, it should work out of the box with no changes to the source code and perform all it's tasks correctly. Further updates might fix bugs or change the output format of some things, but the core should remain the same.

Revising this PR is only a formality as all the code has already been reviewed in the previous PRs.


smorsua (dismissed):
Great job! Couldn't be happier with the outcome.

PR #75, opened 2023-01-26
[PR1] DEV-109 add WebsocketHandle

jmaralo:
The websocket handle unifies all the websocket connections into one by wrapping each message with a type and then sending the previous data with it.

The current implementation encapsulates all the previous components except the logger


src/excel_adapter/internals/excel_reader.go
```diff
@@ -52,7 +49,7 @@ func parseSheet(name string, cols [][]string) models.Sheet {
 	tables := make(map[string]models.Table)
 
 	for name, bound := range findTables(cols) {
-		println(name)
+		fmt.Println(name)
```
lipezaballa, line 52: Delete prints

src/websocket_handle/models/message.go
```diff
@@ -0,0 +1,13 @@
+package models
+
+import "encoding/json"
+
+type Message struct {
+	Kind string          `json:"type"`
```
lipezaballa, line 6: Kind instead of Type?

lipezaballa (approved):

src/connection_transfer/connection_transfer.go
```diff
@@ -36,22 +41,21 @@ func (connectionTransfer *ConnectionTransfer) Update(name string, up bool) {
...
-func (ConnectionTransfer *ConnectionTransfer) Close() {
-	for _, socket := range ConnectionTransfer.sockets {
-		socket.Close()
+func (connectionTransfer *ConnectionTransfer) send() {
+	msg, err := json.Marshal(mapToArray(connectionTransfer.boardStatus))
+	if err != nil {
+		log.Printf("connectionTransfer: send: %s\n", err)
+		return
+	}
+	connectionTransfer.channel <- ws_models.MessageTarget{
```
smorsua, line 53: Create function that returns a `ws_models.MessageTarget` and use it here, in data transfer, etc.

src/websocket_handle/ws_handle.go
```diff
@@ -0,0 +1,98 @@
...
+			case msg := <-output:
+				if len(msg.Target) == 0 {
+					for _, conn := range handle.conns {
+						err := conn.WriteJSON(msg.Msg)
+						if err != nil {
+							handle.Close(conn)
+						}
+					}
+				} else {
+					for _, target := range msg.Target {
```
smorsua, line 46: What does the message target represent?

jmaralo: Where did the message come from or for who the message is, this allows to send the same message to multiple clients or to answer a specific request

src/log_handle/log.go
```diff
@@ -115,28 +122,32 @@ func (logger *LogHandle) createFile(value string) *os.File {
...
 
-func (logger *LogHandle) HandleRequest(w http.ResponseWriter, r *http.Request) {
-	defer r.Body.Close()
-	payload, err := io.ReadAll(r.Body)
-	if err != nil {
-		log.Fatalf("log handle: HandleRequest: %s\n", err)
-	}
+func (logger *LogHandle) listenWS() {
+	for msg := range logger.channel {
+		if logger.logSession == "" || msg.Target[0] == logger.logSession {
```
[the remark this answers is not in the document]
jmaralo: nil is not defined for string, replacing "" with nil throws an error

smorsua (approved):

PR #77, opened 2023-02-09
[PR1] Fixed Sniffer and structures

smorsua:
This PR fixes two bugs caused by the following things we weren't taking into account:

- The sniffer works differently when using localhost than when using tunneling
- The structure table column length varies, its not always the total height


lipezaballa (approved):

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
jmaralo, line 1: elimina este archivo

src/transport_controller/internals/sniffer.go
```diff
@@ -60,26 +61,31 @@ func obtainSource(device string, live bool, config models.Config) *pcap.Handle {
...
 		payload, _, err := sniffer.source.ReadPacketData()
+
 		if err != nil {
 			continue
 		}
 
-		if payload[ETHER_TYPE] != 0x08 || payload[ETHER_TYPE+1] != 0x00 {
-			continue
-		}
+		if os.Getenv("SNIFFER_DEV") == "\\Device\\NPF_Loopback" {
```
jmaralo, line 69: we have to change this, on linux the loopback interface has a different name

jmaralo (approved):
just a few notes, but lgtm

PR #78, opened 2023-02-13
[PR1] DEV-113 Fixed sniffing

smorsua:
This PR fixes the `websocket_handler`, which run into some race conditions. The problem came from iterating and modifying a map at the same time (multiple goroutines). This has know been fixed by annotating the items to be removed and removing them once the iteration is over,


src/websocket_handle/ws_handle.go
```diff
@@ -32,27 +32,33 @@ func RunWSHandle(router *mux.Router, route string, handles map[string]chan model
 
 func (handle *WSHandle) handleTx() {
```
smorsua, line 33: Break into multiple functions

lipezaballa (approved):

jmaralo (approved):

PR #80, opened 2023-02-27
[PR1] DEV-116 Adapted backend to ADE changes

smorsua:
The structure of ADE was changed and it introduces the following changes to the backend:
- The measurements are identified with their ID field, formerly the name field. The current name field is an additional column that was added to ADE. Example: a measurement ID is `airgap_0` and its name is `Airgap 0`.
- The control sections and additional measurements have been removed.

A bug was fixed where the initial value of a measurements sent from the backend was always a string instead of string, int, or bool. Now it sends the appropiate type of value, which allows for the frontend to parse the data correctly.


src/data_transfer/packet_factory.go
```diff
@@ -19,24 +19,24 @@ func NewFactory() PacketFactory {
...
 }
 
-func formatValues(values map[string]any) map[string]string {
-	mapped := make(map[string]string, len(values))
-	for name, val := range values {
-		mapped[name] = fmt.Sprintf("%v", val)
-	}
-	return mapped
-}
+// func formatValues(values map[string]any) map[string]string {
```
jmaralo, line 33: remove comments

jmaralo (approved):

lipezaballa (approved):

PR #82, opened 2023-03-25
[PR 1] DEV-118 Fixes for release 2

jmaralo:
The main changes on this PR are two new abstractions: the vehicle and the boards.

The vehicle abstracts all communication with the vehicle. It exposes two methods: Listen, which takes a channel where new packets will be sent, and SendOrder, which takes an order and sends it to the appropiate board.

The board abstracts concrete logic of a board. Each board intercepts the messages from/to vehicle and applies custom logic based on them (mainly focused toward the BLCU implementation).

A new module, common, has some helper function for some generic tasks. There is also the module channels, which has two custom channel implementation MPSC (multiple producer single consumer) channel and SPMC (single producer multiple consumer) channel, currently no module uses them, but I'll leave them just in case they are needed for something, but probably will be deleted.

Apart from that, some other bugs have been found and fixed, although some known bugs remain, they will be cleaned during the following week, before the app is released.


src/board/mux.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return mux
+}
+
+func (mux *BoardMux) AddBoard(board models.Board) {
+	board.Output(mux.sendOrder)
+	mux.boards[board.Name()] = board
+}
+
+func (mux *BoardMux) AddBoardMapping(idToBoard map[uint16]string) {
```
smorsua, line 37: I would call this inside addBoard because if you add a board, you also want the board mapping.

src/channels/spmc.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (spmc *SPMC[T]) Add(output chan<- T) {
+	spmc.outputMx.Lock()
+	defer spmc.outputMx.Unlock()
+
+	spmc.output = append(spmc.output, output)
+}
+
+func (spmc *SPMC[T]) New(bufSize int) <-chan T {
```
smorsua, line 57: Why did you add the mutex? If its to avoid the concurrent iteration and modification on slice, shouldn't you lock the mutex on the iterations too?

src/channels/mpsc.go
```diff
@@ -0,0 +1,59 @@
...
+			select {
+			case payload := <-channel:
+				mpsc.output <- payload
+			default:
+			}
+		}
+	}
+}
+
+func (mpsc *MPSC[T]) Add(input <-chan T) {
```
smorsua, line 51: You may need a mutex to avoid concurrent iteration and modification of slice

src/cmd/MVP-4/.env
```diff
@@ -1,19 +1,19 @@
-CREDENTIALS=".\\secret.json"
+CREDENTIALS="./secret.json"
```
smorsua, line 1: Does this work on windows?

src/cmd/MVP-4/.env
```diff
@@ -1,19 +1,19 @@
...
 EXCEL_ID="1BEwASubu0el9oQA6PSwVKaNU-Q6gbJ40JR6kgqguKYE"
+#EXCEL_ID="1u0cOaWQI8t7aKfZ4aQCsexKLB3rIPHqX3kDrx0QJZMU"
 EXCEL_PATH="."
 EXCEL_NAME="ade.xlsx"
-LOCAL_ADDRESS="127.0.0.1:6969"
-REMOTE_ADDRESSES="127.0.0.2:5000,127.0.0.3:5000,127.0.0.4:5000,127.0.0.5:5000,127.0.0.6:5000,127.0.0.7:5000,127.0.0.8:5000,127.0.0.9:5000"
+VEHICLE_LADDR="127.0.0.1"
 SNIFFER_LIVE="true"
-SNIFFER_DEV="\\Device\\NPF_Loopback"
+SNIFFER_DEV="lo"
```
smorsua, line 9: What does lo mean?

src/cmd/MVP-4/main.go
```diff
@@ -41,67 +34,52 @@ func main() {
...
+	go vehicle.Listen(vehicleOutput)
+
+	boardMux := board.NewMux(board.WithInput(vehicleOutput), board.WithOutput(vehicle.SendOrder))
+	log.Println("New Mux")
+
+	updateChan := make(chan vehicle_models.Update)
+	go boardMux.Listen(updateChan)
+	log.Println("Mux Listen")
+
+	idToType := make(map[uint16]string)
```
smorsua, line 55: build idToType in separate function

src/cmd/MVP-4/main.go
```diff
@@ -41,67 +34,52 @@ func main() {
...
+			}
 		}
-		raddrs[i] = raddr
 	}
 
 	connectionTransfer, connectionChannel := connection_transfer.New()
+	vehicle.OnConnectionChange(connectionTransfer.Update)
 
-	dataTransfer, dataTransferChannel := data_transfer.New(time.Millisecond * 10)
+	dataTransfer, dataTransferChannel := data_transfer.New(time.Millisecond * 1000 / 30)
```
smorsua, line 75: Make argument frequency instead of period and calculate the period inside the function

src/cmd/MVP-4/main.go
```diff
@@ -110,62 +88,54 @@ func main() {
...
-	httpServerClient := server.Server{Router: mux.NewRouter()}
-	httpServerClient.FileServer("/", filepath.Join(path, "staticClient"))
-
-	go httpServerClient.ListenAndServe(os.Getenv("SERVER_CLIENT_ADDR"))
-
 	interrupt := make(chan os.Signal, 1)
 	signal.Notify(interrupt, os.Interrupt)
 
+	log.Println("backend running!")
 loop:
```
smorsua, line 132: Put loop inside function

src/excel_adapter/excel_adapter.go
```diff
@@ -43,18 +46,30 @@ func getIP(sheet string, document internalModels.Document) string {
 	panic(fmt.Sprintf("excel adapter: getIP: Missing board %s IP\n", sheet))
 }
 
-func AddExpandedPackets(document internalModels.Document, objects ...models.FromBoards) {
-	globalInfo := models.GlobalInfo{BoardToIP: getBoardToIPs(document), UnitToOperations: make(map[string]string)}
+func Update(document internalModels.Document, objects ...models.FromDocument) {
+	globalInfo := getGlobalInfo(document)
```
smorsua, line 50: Put in separate `addGlobalInfo` function

src/excel_adapter/excel_adapter.go
```diff
@@ -43,18 +46,30 @@ func getIP(sheet string, document internalModels.Document) string {
...
+	for _, object := range objects {
+		object.AddGlobal(globalInfo)
+	}
 
 	for _, board := range getBoards(document) {
```
smorsua, line 55: Put in separate `addPackets` function

src/excel_adapter/internals/excel_reader.go
```diff
@@ -26,8 +26,8 @@ func parseSheets(file *excelize.File) (models.Sheet, map[string]models.Sheet) {
 	sheetMap := file.GetSheetMap()
 	for _, name := range sheetMap {
 		cols := getSheetCols(file, name)
-		if strings.HasPrefix(name, BOARD_SHEET_PREFIX) {
-			boardSheets[strings.TrimPrefix(name, BOARD_SHEET_PREFIX)] = parseSheet(name, cols)
+		if !strings.HasPrefix(name, GLOBAL_SHEET_PREFIX) {
+			boardSheets[strings.TrimPrefix(name, GLOBAL_SHEET_PREFIX)] = parseSheet(name, cols)
```
smorsua, line 30: Is this correct? You are removing the `GLOBAL_SHEET_PREFIX`, maybe you intended to remove the `BOARD_SHEET_PREFIX`

src/excel_adapter/models/control_sections.go
smorsua, line 1: Remove this file, is deprecated

src/message_transfer/message_transfer.go
```diff
@@ -26,15 +26,15 @@ func (messageTransfer *MessageTransfer) Broadcast(update dataTransferModels.Pack
...
 	var message models.Message
-	if msg, ok := update.Values["warning"]; ok {
+	if msg, ok := update.Fields["warning"]; ok {
 		message = models.Message{
 			ID:          update.ID,
 			Description: msg.(string),
 			Type:        "warning",
 		}
-	} else if msg, ok = update.Values["fault"]; ok {
+	} else if msg, ok = update.Fields["fault"]; ok {
```
smorsua, line 37: Add else with an error message of incorrect message type.

src/packet_parser/packet_parser.go
```diff
@@ -44,8 +46,8 @@ func (parser *PacketParser) AddPacket(globalInfo excelAdapterModels.GlobalInfo,
 	parser.descriptors[uint16(id)] = valueDescriptors
 }
 
-func NewPacketParser() PacketParser {
-	return PacketParser{
+func NewPacketParser() *PacketParser {
```
smorsua, line 49: Why pointer?

src/vehicle/models/pod_data.go
```diff
@@ -12,38 +12,40 @@ type PodData struct {
 	Boards map[string]Board `json:"boards"`
 }
 
-func NewPodData() PodData {
-	return PodData{
+func NewPodData() *PodData {
```
smorsua, line 15: Why pointer?

src/vehicle/vehicle.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/HyperloopUPV-H8/Backend-H8/pipe"
+	"github.com/HyperloopUPV-H8/Backend-H8/sniffer"
+	"github.com/HyperloopUPV-H8/Backend-H8/unit_converter"
+	"github.com/HyperloopUPV-H8/Backend-H8/vehicle/internals"
+	"github.com/HyperloopUPV-H8/Backend-H8/vehicle/models"
+)
+
+type Vehicle struct {
+	sniffer          *sniffer.Sniffer
+	parser           *packet_parser.PacketParser
```
smorsua, line 17: Why are they pointers?

src/common/io.go
```diff
@@ -0,0 +1,16 @@
+package common
+
+import "io"
+
+func WriteAll(writer io.Writer, data []byte) (n int, err error) {
```
smorsua, line 5: Where do you use this?

smorsua, line 5: Where do you use this?

smorsua, line 5: Call the first returned value written instead of n because you have another variable inside the function called n and it's confusing

jmaralo: Inside Vehicle.SendOrder to ensure every byte is written

src/vehicle/models/pod_data.go
jmaralo: because AddGlobal and AddPacket work on a pointer to the type

src/packet_parser/packet_parser.go
jmaralo: AddGlobal and AddPacket work on a reference

src/vehicle/vehicle.go
jmaralo: because AddGlobal and AddPacket work on a pointer to the type

src/excel_adapter/internals/excel_reader.go
jmaralo: This works better, there is only one global sheet and the rest are all boards, this way the names on the excel adapter for boards are just the board name and then the sheet with the global info is the one that has a prefix

src/cmd/MVP-4/main.go
jmaralo: I don't see the use of putting this on a function, mainly because this will be removed before release. This is just for debugging and we should use a better method

src/cmd/MVP-4/main.go
jmaralo: I prefer this approach. A duration has units (seconds, milliseconds, etc.), if we use the frequency, the user must know weather it is in FPS or some other random measurement we decided. I'll instead make a function that calculates the duration from the FPS

src/board/mux.go
jmaralo: Yeah, but maybe you dont have that information when you add the board logic or you have it before. Or maybe you want to change the board logic without changing the mapping. This approach is more flexible

src/channels/mpsc.go
jmaralo: That doesnt happen because you are appending to the slice. When iterating in a for loop using range, range is evaluated before iterating, thus, this change doesn't take effect until the next iteration without causing any problems.

src/channels/spmc.go
jmaralo: forgot to remove it

src/cmd/MVP-4/.env
jmaralo: I don't think so, making this compatible with multiple OS will be tricky

src/cmd/MVP-4/.env
jmaralo: is the name of the interface on my machine

Losina24 (approved):
It is a good practice to make small PRs, keep in mind that the main objective of this methodology is to make it easier for other developers to review your code.

src/cmd/MVP-4/.env
```diff
@@ -1,19 +1,19 @@
-CREDENTIALS=".\\secret.json"
+CREDENTIALS="./secret.json"
 #EXCEL_ID="1ppqYGlGTXGb-PKOcMesyvz12pLuD1XoPAJ0N_kpRJtg"
 EXCEL_ID="1BEwASubu0el9oQA6PSwVKaNU-Q6gbJ40JR6kgqguKYE"
+#EXCEL_ID="1u0cOaWQI8t7aKfZ4aQCsexKLB3rIPHqX3kDrx0QJZMU"
```
Losina24, line 4: If u r not using it, then remove it. If u need to recover it in any moment u can retrieve the value from git

src/message_transfer/message_transfer.go
Losina24: And by the way, change the else if to a non-indented if (there are better ways to do this but it's not worth implementing at this point)

src/vehicle/models/pod_data.go
```diff
@@ -64,17 +66,44 @@ func getMeasurements(values []excelAdapterModels.Value) map[string]Measurement {
...
+	strRange := strings.Split(strings.TrimSuffix(strings.TrimPrefix(strings.Replace(literal, " ", "", -1), "["), "]"), ",")
+
+	if len(strRange) != 2 {
 		log.Fatalf("pod data: parseRange: invalid range %s\n", literal)
 	}
 
-	return [2]string{split[0], split[1]}
+	numRange := make([]*float64, 0)
+
+	if strRange[0] != "" {
```
Losina24, line 82: it can be refactored a lot by extracting to methods and using guard clauses

src/cmd/MVP-4/main.go
smorsua: frecuency and fps are the same

src/vehicle/models/pod_data.go
jmaralo: I just moved this file, I leave those changes for the next PR

src/cmd/MVP-4/main.go
jmaralo: frequency is the inverse of the period, but by passing a frequency you are assuming it is in s^-1, this way the user can say if it is in seconds, milliseconds, etc. It has more meaning

src/excel_adapter/excel_adapter.go
jmaralo: We are splitting the code too much I think by doing so

src/excel_adapter/excel_adapter.go
jmaralo: Same as before, I don't like the idea of having one function for each for loop

smorsua (approved):
Nice work

src/cmd/MVP-4/main.go
```diff
@@ -110,62 +73,78 @@ func main() {
...
+		measurements_loop:
+			for msr := range pkt.Measurements {
+				if msr == "warning" {
+					idToType[pkt.ID] = "warning"
+					break measurements_loop
+				} else if msr == "fault" {
+					idToType[pkt.ID] = "fault"
+					break measurements_loop
+				}
+			}
```
lipezaballa, line 146: may be too much for a function

Losina24: Of course, use breaks is a clear symptom. At least 3 methods should be extracted from here

PR #83, opened 2023-03-28
[PR 1] DEV-121 improve WebSocketBroker implementation

jmaralo:
The previous implementation of the WebSocketBroker used a single channel, both to read and send messages, for each handler. That could cause some messages to be received by the same sender.

Now we have Handlers, they implement an interface and need to be registered on the WebSocketBroker before use.

The handlers are DataTransfer, MessageTransfer, OrderTransfer, ConnectionTransfer and LogHandle. Their implementation has also be changed to a singleton, as it doesn't make sense for multiple instances of them to exist at the same time.

With the new implementation:

- Multiple handlers can listen for the same topic
- Two different channels for sending and receiving
- No need to build a `ws_models.MessageTarget` outside `web_socket_broker`

**NOTE**: Some values are hard coded into constants, this is intended to change in future updates


src/cmd/MVP-4/main.go
```diff
@@ -46,46 +45,40 @@ func main() {
...
-	})
+	vehicle.OnConnectionChange(connectionTransfer.Update)
 
-	go func(msgChannel chan models.MessageTarget) {
+	go func() {
 		for update := range updateChan {
-			logger.Update(update.Fields)
-			if idToType[update.ID] == "data" {
+			logger.Update(update)
+			if msg, err := message_transfer_models.MessageFromUpdate(update); err != nil {
```
smorsua, line 72: Make a separate function that handles detecting the type of update instead of returning error from MessageFromUpdate

smorsua (approved):

src/cmd/MVP-4/.env
```diff
@@ -1,7 +1,5 @@
 CREDENTIALS="./secret.json"
-#EXCEL_ID="1ppqYGlGTXGb-PKOcMesyvz12pLuD1XoPAJ0N_kpRJtg"
```
Losina24, line 2: If u r not using it better remove it

src/websocket_broker/ws_handle.go
```diff
@@ -0,0 +1,165 @@
+package websocket_broker
+
```
Losina24, line 2: this file can (and should) get slimmer

src/websocket_broker/ws_handle.go
```diff
@@ -0,0 +1,165 @@
...
+type WebSocketBroker struct {
+	handlers   map[string][]models.MessageHandler
+	handlersMx *sync.Mutex
+	clients    map[string]*websocket.Conn
+	clientsMx  *sync.Mutex
+}
+
+func (broker *WebSocketBroker) HandleConn(writter http.ResponseWriter, request *http.Request) {
+	defer request.Body.Close()
```
Losina24, line 46: for example, move this logic into a new WebSockerConnectionHandler file

jmaralo: The code is better understood if everything is on the same file, and the file is not too long

Losina24: Precisely, having everything in one file makes the code more difficult to read, difficult to understand and prevents it from being scalable.

Like everything I usually comment on here, it does not mean that the solution you choose is wrong, of course, I simply leave tips that will help to make a more professional and robust code.

I can buy the argument of "it's not worth changing this" because in the end you know how much time you have and many times I'm "picky" with this kind of thing, but my goal is that if you leave the code like that, it's because you've decided to do it consciously and considering the cons, and not because you have not known how to do it in another way.

That being said, it is true that a 150-200 line file is not excessively large, but in this case, there is a lot of extractable logic being written here. The fact that a method is extracted does not mean that this method no longer exists in this file, the inner logic is simply moved to an external file and you access it from the same current function of this file . This makes it easier to read because, for example, if you are looking for a bug in the HandleConn function, your file will have little content and you will be able to see very easily that the logic with which you have problems is in its own file that takes care of that. So the rest of the logic of the other functions does not interfere with your true objective.

Losina24: and a clear symptom of this is that the code has many indentations and you make use of breaks and empty returns in the middle of loops

PR #87, opened 2023-04-04
[PR 1] BLCU implementation

jmaralo:
This PR implements the upload and download logic of the BLCU. It uses the previously implemented board abstraction to get notified about acknowledgements.

To upload/download code from the BLCU, first we need to send the corresponding order to the board, then we wait for an acknowledgement. Once we get the ack, we can begin writing/reading the file from the TFTP server.


src/board/boards/blcu/blcu.go
```diff
@@ -0,0 +1,114 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"strconv"
+
+	excel_models "github.com/HyperloopUPV-H8/Backend-H8/excel_adapter/models"
+	"github.com/HyperloopUPV-H8/Backend-H8/vehicle/models"
+)
+
+// TODO: Find a better way to store these values (.env?)
```
smorsua, line 13: In the toml

src/board/boards/blcu/download.go
```diff
@@ -0,0 +1,86 @@
...
+	// TODO: remove hardcoded timeout
+	if _, err := common.ReadTimeout(blcu.ackChannel, time.Minute); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getDownloadOrder(board string) models.Order {
+	return models.Order{
```
smorsua, line 48: Rename this function and getUploadOrder to createDownloadOrder and createUploadOrder

smorsua (changes requested):
Very clean, get back to me when you apply the changes

src/board/boards/blcu/blcu.go
```diff
@@ -0,0 +1,114 @@
...
+	case os.Getenv("BLCU_DOWNLOAD_TOPIC"):
+		if file, err := blcu.handleDownload(payload); err != nil {
+			blcu.notifyDownloadFailure()
+		} else {
+			blcu.notifyDownloadSuccess(file)
+		}
+	}
+}
+
+func (blcu *BLCU) SetSendMessage(sendMessage func(topic string, payload any, targets ...string) error) {
```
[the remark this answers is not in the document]
jmaralo: this is a method required for an interface on websocket broker, we don't know what this function is before creating the websocket handler and that function is set by websocketBroker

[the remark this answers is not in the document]
jmaralo: this function is necessary for an interface required by WebsocketBroker to register a new handler, it is called by WebsocketBroker to provide the function that can be used to send a message to the frontend

src/board/boards/blcu/upload.go
```diff
@@ -0,0 +1,88 @@
...
+	_, err = sender.ReadFrom(reader)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// the topic BLCU_STATE_TOPIC expects a number between 0 and 100, the idea is in the future to inform about the percentage of the file uploaded
+func (blcu *BLCU) notifyUploadFailure() {
```
lipezaballa, line 80: leave it as a TODO to remember to complete it

lipezaballa (approved):

src/board/boards/blcu/upload.go
jmaralo: we dont know if it will be possible right now, so we are just leaving it like this, just knowing when the code is uploaded is ok for now

src/board/boards/blcu/blcu.go
```diff
@@ -0,0 +1,128 @@
...
+	inputChannel chan models.Update
+	ackChannel   chan struct{}
+
+	sendOrder   func(models.Order) error
+	sendMessage func(topic string, payload any, targets ...string) error
+
+	trace zerolog.Logger
+}
+
+func NewBLCU() *BLCU {
```
smorsua, line 42: Why do you return a poiter?

jmaralo: It is necessary because the interface methods require a pointer

lipezaballa (approved):

smorsua (approved):

PR #88, opened 2023-04-05
[PR1] DEV-131 Replaced .env with config.toml and open browser

smorsua:
This PR changes the .env for a .toml file. We have done this change because the env was getting to big and doesn't allow for any nesting.

Additionally, the frontend is opened in the browser when the backend is run. Finally, the cmd folder has been removed and its contents are at the src folder. I didn't change the trace.go file or the logger.go file so it can be safely ignored.


src/config.toml
```diff
@@ -0,0 +1,51 @@
...
+file_server_path = "static" # Change name to front build path
+endpoints = { pod_data = "podDataStructure", order_data = "orderDataStructure", websocket = "", file_server = "/" }
+
+[vehicle]
+tcp_client_tag = "TCP_CLIENT"
+tcp_server_tag = "TCP_SERVER"
+udp_tag = "UDP"
+network = { mtu = 1500, sniffer_interface = "\\Device\\NPF_Loopback" }
+
+# [excel]
```
jmaralo, line 13: does this work without this?

src/connection_transfer/connection_transfer.go
```diff
@@ -15,10 +14,21 @@ const (
 	CONNECTION_TRANSFER_HANDLER_NAME = "connectionTransfer"
 )
 
+type ConnectionTransferConfig struct {
```
jmaralo, line 17: move this to models or to a separate `config.go` file

src/connection_transfer/connection_transfer.go
```diff
@@ -15,10 +14,21 @@ const (
...
+	UpdateTopic string `toml:"update_topic"`
+}
+
 var (
+	connectionTransferConfig = ConnectionTransferConfig{
```
jmaralo, line 22: instead of doing this for the configuration of every component, make the whole configuration be a singleton and provide a getter for the component configuration that panics if the configuration is not initialized

src/connection_transfer/connection_transfer.go
```diff
@@ -15,10 +14,21 @@ const (
...
+		UpdateTopic: "connection/update",
+	}
 	connectionTransfer *ConnectionTransfer
 )
 
+func SetConfig(config ConnectionTransferConfig) {
```
jmaralo, line 28: you don't need one of these functions if you make config global, it also ensures you dont forget to call it

src/config.toml
smorsua: Yes

src/data_transfer/data_transfer.go
```diff
@@ -17,10 +15,26 @@ const (
 	DATA_TRANSFER_HANDLER_NAME = "dataTransfer"
 )
 
+type DataTransferTopics struct {
```
jmaralo, line 18: same as on ConnectionTransfer, put config in other file, store all the config in DataTransfer and get from global config

src/excel_adapter/internals/excel_reader.go
```diff
@@ -2,32 +2,43 @@ package internals
...
+	TablePrefix       string `toml:"table_prefix"`
+	AddressTable      string `toml:"address_table"`
+	BackendEntryKey   string `toml:"backend_entry_key"`
+	UnitsTable        string `toml:"units_table"`
+	PortsTable        string `toml:"ports_table"`
+	IdsTable          string `toml:"ids_table"`
+}
+
+func GetDocument(file *excelize.File, config ParseConfig) models.Document {
+	infoSheet, boardSheets := parseSheets(file, config)
```
jmaralo, line 23: get this from global config

src/excel_adapter/internals/excel_reader.go
```diff
@@ -2,32 +2,43 @@ package internals
...
-	"os"
 	"strings"
 
 	"github.com/HyperloopUPV-H8/Backend-H8/excel_adapter/internals/models"
 	"github.com/xuri/excelize/v2"
 )
 
-func GetDocument(file *excelize.File) models.Document {
-	infoSheet, boardSheets := parseSheets(file)
+type ParseConfig struct {
```
jmaralo, line 11: keep in a separate file

src/excel_adapter/internals/excel_retriever.go
```diff
@@ -13,33 +13,40 @@ import (
 
 const SHEETS_MIME_TYPE = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
 
-func DownloadFile(id string, path string, name string) error {
-	trace.Trace().Str("id", id).Str("path", path).Str("name", name).Msg("download file")
-	client, errClient := getClient()
+type DownloadConfig struct {
```
jmaralo, line 16: define in a separate file

src/excel_adapter/internals/excel_retriever.go
```diff
@@ -13,33 +13,40 @@ import (
...
+	Id          string
+	Path        string
+	Name        string
+	Credentials string
+}
+
+func DownloadFile(config DownloadConfig) error {
```
jmaralo, line 23: get config from global config file

src/excel_adapter/excel_adapter.go
```diff
@@ -11,35 +10,65 @@ import (
 	"github.com/xuri/excelize/v2"
 )
 
-func FetchDocument(id string, path string, name string) internalModels.Document {
-	trace.Info().Str("id", id).Str("path", path).Str("name", name).Msg("fetch document")
+type ExcelAdapterConfig struct {
```
jmaralo, line 13: place on a different file

src/excel_adapter/excel_adapter.go
```diff
@@ -11,35 +10,65 @@ import (
...
+
+type ExcelAdapter struct {
+	config     ExcelAdapterConfig
+	boards     map[string]models.Board
+	globalInfo models.GlobalInfo
+	document   internalModels.Document
+}
+
+func NewExcelAdapter(config ExcelAdapterConfig) ExcelAdapter {
+	//TODO: poner config del download nesteado en el TOML
```
jmaralo, line 26: is this comment already fixed?

src/message_transfer/message_transfer.go
```diff
@@ -29,19 +39,21 @@ func Get() *MessageTransfer {
...
 	trace.Info().Msg("init message transfer")
 	messageTransfer = &MessageTransfer{
+		updateTopic: messageTransferConfig.updateTopic,
 		sendMessage: defaultSendMessage,
 		trace:       trace.With().Str("component", MESSAGE_TRANSFER_HANDLER_NAME).Logger(),
 	}
 }
 
 type MessageTransfer struct {
+	updateTopic string
```
jmaralo, line 49: store config struct

src/server/server.go
```diff
@@ -9,6 +9,18 @@ import (
 	trace "github.com/rs/zerolog/log"
 )
 
+type ServerConfig struct {
```
jmaralo, line 12: different file

jmaralo: where is this struct being used?

src/sniffer/sniffer.go
```diff
@@ -19,10 +16,15 @@ type Sniffer struct {
 	trace  zerolog.Logger
 }
 
-func New(dev string, filter string) (*Sniffer, error) {
+type SnifferConfig struct {
```
jmaralo, line 19: different file

jmaralo, line 19: different file

src/vehicle/builder.go
```diff
@@ -32,7 +32,17 @@ type Builder struct {
 	trace            zerolog.Logger
 }
 
-func NewBuilder() *Builder {
+type BuilderConfig struct {
```
jmaralo, line 35: different file

src/config.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/HyperloopUPV-H8/Backend-H8/connection_transfer"
+	"github.com/HyperloopUPV-H8/Backend-H8/data_transfer"
+	"github.com/HyperloopUPV-H8/Backend-H8/excel_adapter"
+	"github.com/HyperloopUPV-H8/Backend-H8/log_handle"
+	"github.com/HyperloopUPV-H8/Backend-H8/message_transfer"
+	"github.com/HyperloopUPV-H8/Backend-H8/server"
+	"github.com/HyperloopUPV-H8/Backend-H8/vehicle"
+)
+
+type Config struct {
```
jmaralo, line 13: have this as a singleton that is initialized from main, then provide getter methods to access module configuration.

This way you dont need to pass the config to all methods, it can be accessed globally.

src/config.toml
```diff
@@ -0,0 +1,50 @@
...
+tcp_server_tag = "TCP_SERVER"
+udp_tag = "UDP"
+network = { mtu = 1500, sniffer_interface = "\\Device\\NPF_Loopback" }
+
+[excel.download]
+id = "1BEwASubu0el9oQA6PSwVKaNU-Q6gbJ40JR6kgqguKYE"
+credentials = "./secret.json"
+name = "ade.xlsx"
+path = "."
+[excel.parse]
```
jmaralo, line 18: add white line in between, just so it is cleaner :P

src/main.go
```diff
@@ -58,17 +61,25 @@ func main() {
 	// Communication with front-end
 	websocketBroker := websocket_broker.Get()
 
+	connection_transfer.SetConfig(config.Connections)
```
jmaralo, line 64: you dont need to pass the config to every single method if the config is global

src/main.go
```diff
@@ -31,19 +33,20 @@ var traceFile = flag.String("log", "trace.json", "set the trace log file")
...
 	runtime.GOMAXPROCS(runtime.NumCPU())
-	godotenv.Load(".env")
-
 	flag.Parse()
+
 	traceFile := initTrace(*traceLevel, *traceFile)
 	defer traceFile.Close()
 
-	document := excel_adapter.FetchDocument(os.Getenv("EXCEL_ADAPTER_ID"), os.Getenv("EXCEL_ADAPTER_PATH"), os.Getenv("EXCEL_ADAPTER_NAME"))
+	config := getConfig()
```
jmaralo, line 41: put this method in config as `InitConfig`

src/main.go
```diff
@@ -137,3 +149,20 @@ func getIdToType(podData *vehicle_models.PodData) map[uint16]string {
 	}
 	return idToType
 }
+
+func getConfig() Config {
+	configFile, fileErr := os.ReadFile("./config.toml")
```
jmaralo, line 154: instead of hardcoding the config file, add a flag to get it.

A better way would be to have this method get as a parameter the config file path

jmaralo (changes requested):
Looks great, but I think there are better ways to store and propagate the configuration.

Instead of propagating the global config through functions arguments, make some getters for the config and store the config as a global variable inside a module, then, when you need to access the configuration call the getter from inside the function.

This is the approach I would take, but I want to hear your opinion.

src/excel_adapter/excel_adapter.go
```diff
@@ -126,6 +126,9 @@ func getInfoTableToMap(tableName string, document internalModels.Document) map[s
 	}
 
 	for _, row := range table.Rows {
+		if row[0] == "Backend" {
```
jmaralo, line 129: should add a TODO to remove this workaround

jmaralo (approved):
lgtm, I would place the configuration structs definition on other files, but this also works

lipezaballa (approved):

PR #90, opened 2023-04-14
[PR 1] DEV-139 Removed SetGlobal and SetPacket

smorsua:
This PR removes SetGlobal and SetPacket and instead makes all required data a parameter in the constructor of each module. This should make debugging easier as objects are completely built from the start.

### DIFF
--- a/src/board/boards/blcu/download.go
+++ b/src/board/boards/blcu/download.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"io"
-	"os"
 	"time"
 
 	"github.com/HyperloopUPV-H8/Backend-H8/common"
@@ -87,12 +86,10 @@ type fileResponsePayload struct {
 
 func (blcu *BLCU) notifyDownloadFailure() {
 	blcu.trace.Warn().Msg("Download failed")
-	//FIXME: change to toml
-	blcu.sendMessage(os.Getenv("BLCU_DOWNLOAD_TOPIC"), fileResponsePayload{IsSuccess: false, File: nil})
+	blcu.sendMessage(blcu.downloadTopic, fileResponsePayload{IsSuccess: false, File: nil})
 }
 
 func (blcu *BLCU) notifyDownloadSuccess(bytes []byte) {
-	//FIXME: change to toml
 	blcu.trace.Info().Msg("Download success")
-	blcu.sendMessage(os.Getenv("BLCU_DOWNLOAD_TOPIC"), fileResponsePayload{IsSuccess: true, File: bytes})
+	blcu.sendMessage(blcu.downloadTopic, fileResponsePayload{IsSuccess: true, File: bytes})
 }
--- a/src/board/boards/blcu/download.go
+++ b/src/board/boards/blcu/download.go
@@ -86,10 +86,10 @@ type fileResponsePayload struct {
 
 func (blcu *BLCU) notifyDownloadFailure() {
 	blcu.trace.Warn().Msg("Download failed")
-	blcu.sendMessage(blcu.downloadTopic, fileResponsePayload{IsSuccess: false, File: nil})
+	blcu.sendMessage(blcu.config.Topics.Download, fileResponsePayload{IsSuccess: false, File: nil})
 }
 
 func (blcu *BLCU) notifyDownloadSuccess(bytes []byte) {
 	blcu.trace.Info().Msg("Download success")
-	blcu.sendMessage(blcu.downloadTopic, fileResponsePayload{IsSuccess: true, File: bytes})
+	blcu.sendMessage(blcu.config.Topics.Download, fileResponsePayload{IsSuccess: true, File: bytes})
 }
--- a/src/board/boards/blcu/download.go
+++ b/src/board/boards/blcu/download.go
@@ -87,10 +87,12 @@ type fileResponsePayload struct {
 
 func (blcu *BLCU) notifyDownloadFailure() {
 	blcu.trace.Warn().Msg("Download failed")
+	//FIXME: change to toml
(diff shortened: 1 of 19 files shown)


src/board/boards/blcu/download.go
```diff
@@ -87,10 +87,12 @@ type fileResponsePayload struct {
 
 func (blcu *BLCU) notifyDownloadFailure() {
 	blcu.trace.Warn().Msg("Download failed")
+	//FIXME: change to toml
 	blcu.sendMessage(os.Getenv("BLCU_DOWNLOAD_TOPIC"), fileResponsePayload{IsSuccess: false, File: nil})
```
jmaralo, line 91: you can fix this now, it seems I forgot to change these functions, the blcu has a config field which should have the topics

src/config.toml
```diff
@@ -1,18 +1,15 @@
...
-
-[server.endpoints]
-pod_data = "/podDataStructure"
-order_data = "/orderStructures"
-websocket = "/backend"
-file_server = "/"
+#address = "192.168.0.9:4000"
+address = "127.0.0.9:4000"
+file_server_path = "static" # Change name to front build path
+endpoints = { pod_data = "/podDataStructure", order_data = "/orderDataStructure", websocket = "/backend", file_server = "/" }
```
jmaralo, line 6: I find this inline table too long, better make it its own table like before
```suggestion
[server.endpoints]
pod_data = "/podDataStructure"
order_data = "/orderStructures"
websocket = "/backend"
file_server = "/"
```

src/connection_transfer/connection_transfer.go
```diff
@@ -18,7 +18,7 @@ const (
 type ConnectionTransfer struct {
 	writeMx     *sync.Mutex
 	boardStatus map[string]models.Connection
-	sendMessage func(topic string, payload any, target ...string) error
+	SendMessage func(topic string, payload any, target ...string) error
```
jmaralo, line 21: why is this public?

src/excel_adapter/excel_adapter.go
```diff
@@ -15,29 +15,20 @@ type ExcelAdapterConfig struct {
...
-type ExcelAdapter struct {
-	config     ExcelAdapterConfig
-	boards     map[string]models.Board
-	globalInfo models.GlobalInfo
-	document   internalModels.Document
-}
-
-func NewExcelAdapter(config ExcelAdapterConfig) ExcelAdapter {
-	//TODO: poner config del download nesteado en el TOML
+func FetchBoardsAndGlobalInfo(config ExcelAdapterConfig) (map[string]models.Board, models.GlobalInfo) {
```
jmaralo, line 18: just to know your opinion, what about splitting this method into two different ones and then calling both in main?

src/excel_adapter/excel_adapter.go
```diff
@@ -15,29 +15,20 @@ type ExcelAdapterConfig struct {
...
-		globalInfo: getGlobalInfo(document, GlobalInfoConfig{
-			AddressTable:    config.Parse.AddressTable,
-			BackendEntryKey: config.Parse.BackendEntryKey,
-			UnitsTable:      config.Parse.UnitsTable,
-			PortsTable:      config.Parse.PortsTable,
-			IdsTable:        config.Parse.IdsTable,
-		}),
-		document: document,
-	}
+	globalInfo := getGlobalInfo(document, GlobalInfoConfig{
```
jmaralo, line 21: this is just me being picky, let me know your opinion, what about changing the parser config struct to be the three other fields and then this struct and in the config you have:

```toml
[excel.parse]
global_sheet_prefix = "GLOBAL "
board_sheet_prefix = "BOARD "
table_prefix = "[TABLE] "
global.address_table = "addresses"
global.backend_entry_key = "Backend"
global.units_table = "units"
global.ports_table = "ports"
global.ids_table = "ids"
```

(or changing global with any other identifier)

src/board/boards/blcu/download.go
smorsua: Okay. The blcu still has the AddPacket function which will be removed, but I'll do it in the next PR because getting the ACK id is related to some changes that I'm currently working on.

jmaralo (changes requested):
Just a few things to improve and some questions

src/excel_adapter/excel_adapter.go
smorsua: I will do it in the next PR, which introduces some changes to the excelAdapter

src/connection_transfer/connection_transfer.go
smorsua: It shouldn't, thanks for the heads up

src/excel_adapter/excel_adapter.go
smorsua: Sure thing, i'll introduce this change in the next PR

src/config.toml
smorsua: I'll introduce this change in the next PR

src/unit_converter/unit_converter.go
```diff
@@ -10,40 +10,34 @@ import (
...
-	for _, val := range packet.Values {
-		var ops string
-		if converter.Kind == "pod" && val.PodUnits != "" {
-			ops = getCustomOrGlobalOperations(val.PodUnits, converter.unitOperations)
-		} else if converter.Kind == "display" && val.DisplayUnits != "" {
-			ops = getCustomOrGlobalOperations(val.DisplayUnits, converter.unitOperations)
-		} else {
-			converter.trace.Trace().Str("id", val.ID).Msg("no unit conversion")
-			continue
+	for _, board := range boards {
```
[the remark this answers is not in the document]
smorsua: There is no logic between the loops so it's not necessary to split it

jmaralo (approved):

PR #92, opened 2023-04-16
[PR1] DEV-141 add moving average

jmaralo:
_DO NOT MERGE BEFORE DEV-140, this branch originates from DEV-140 as the changes there were necessary to test with the front end_

Previously values were sent raw and were difficult to visualize in Ethernet View, this implements a moving average for numeric values and the cycle time so it is easier to see when just a number is displayed.

The moving average takes the N (currently set to 100) last values and computes the mean. Ideally, in the future N will be calculated based on the cycle time itself, so the average for less frequent packets is closer to the real value.

There are also some fixes to the unit conversion logic that wasn't working as expected.

### DIFF
--- a/src/message_parser/models/protection_message.go
+++ b/src/message_parser/models/protection_message.go
@@ -19,35 +19,40 @@ type ProtectionMessage struct {
 
 type ProtectionTimestamp struct {
 	Counter uint16 `json:"counter"`
-	Second  uint8  `json:"second"`
-	Minute  uint8  `json:"minute"`
-	Hour    uint8  `json:"hour"`
+	Seconds uint8  `json:"seconds"`
+	Minutes uint8  `json:"minutes"`
+	Hours   uint8  `json:"hours"`
 	Day     uint8  `json:"day"`
 	Month   uint8  `json:"month"`
 	Year    uint8  `json:"year"`
 }
 
 type OutOfBoundsViolation struct {
+	Kind string     `json:"kind"`
 	Want [2]float64 `json:"want"`
 	Got  float64    `json:"got"`
 }
 
 type UpperBoundViolation struct {
+	Kind string  `json:"kind"`
 	Want float64 `json:"want"`
 	Got  float64 `json:"got"`
 }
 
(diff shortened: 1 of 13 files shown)


src/message_parser/models/protection_message.go
jmaralo, line 1: Please ignore this file, it seems I screwed up when merging some changes from DEV-140 and this file shouldn't appear as modified

src/vehicle/vehicle.go
```diff
@@ -57,7 +57,12 @@ func (vehicle *Vehicle) Listen(updateChan chan<- models.Update, messagesChan cha
 	}()
 	go func() {
 		for raw := range vehicle.messageChan {
-			messagesChan <- vehicle.messageParser.Parse(raw)
+			msg, err := vehicle.messageParser.Parse(raw)
```
jmaralo, line 60: This is also from DEV-140, ignore it

src/vehicle/internals/update_factory.go
```diff
@@ -4,43 +4,101 @@ import (
...
 }
 
 func NewFactory() UpdateFactory {
 	trace.Info().Msg("new update factory")
 	return UpdateFactory{
-		count:     make(map[uint16]uint64),
-		timestamp: make(map[uint16]uint64),
-		trace:     trace.With().Str("component", "updateFactory").Logger(),
+		count:        make(map[uint16]uint64),
+		cycleTimeAvg: make(map[uint16]*common.MovingAverage[uint64]),
```
smorsua, line 27: Why are the moving averages pointers instead of values?

smorsua (commented):
Nice implementation, just a couple of details

src/vehicle/internals/update_factory.go
jmaralo: to prevent copies, because methods of updateFactory use it as a value instead of a reference, which can create copies of the value

lipezaballa (approved):

smorsua (approved):

PR #95, opened 2023-04-17
[PR1] DEV-146 implement bit array

jmaralo:
This encodes the enabled and disabled fields in a bit array that is sent with the order so the boards know what fields should be taken into consideration

### DIFF
--- a/src/packet_parser/packet_parser.go
+++ b/src/packet_parser/packet_parser.go
@@ -3,6 +3,7 @@ package packet_parser
 import (
 	"bytes"
 	"io"
+	"math"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


smorsua (approved):

Losina24 (approved):

src/packet_parser/packet_parser.go
```diff
@@ -146,3 +147,33 @@ func (parser PacketParser) encodeValue(desc models.ValueDescriptor, value any, w
...
+
+		if enableMap[valueDescriptor.ID] {
+			value = 1
+		} else {
+			value = 0
+		}
+
+		bitArray[i/8] = setBit(bitArray[i/8], i%8, value)
+		i++
+	}
```
Losina24, line 167:
```go
        for _, valueDescriptor := range parser.descriptors[id] {
		bitArray[i/8] = setBit(bitArray[i/8], i%8, getBitValue(id))
		i++
	}

        func getBitValue(id uint16) {
               if enableMap[valueDescriptor.ID] { return 1 }
               return 0
        }
```

PR #96, opened 2023-04-19
[PR1] DEV-147 improve packet parser structure

jmaralo:
This is still a work in progress as some parts of the code are really dirty, most parts aren't instrumented properly and further testing must be conducted.

The idea is to provide a more general parser which will decode all traffic and provide an interface with good types (use less `interface{}` or `any`)

`packet.go` is the folder where almost all of the changes have been made, the rest are just updates to the existing code to use the new interface.

Some things have been pulled from other branches as they were needed to further develop and test this implementation.

Once the code is more mature, the idea is to split the code in multiple PRs so it is easier to review.


Losina24 (approved):
66 files -> autoapprove

Losina24 (approved):

jmaralo:
obsolete

PR #98, opened 2023-04-24
[PR1] DEV-152 Vehicle Refactor

smorsua:
The Measurement model has been split into NumericMeasurement, EnumMeasurement and BooleanMeasurement, so all of the following changes are to accommodate these new models. We store them as any differentiate them by using a type switch.


lipezaballa (changes requested):

src/vehicle/constructor.go
```diff
@@ -0,0 +1,251 @@
...
+		onConnectionChange: args.OnConnectionChange,
+		trace:              vehicleTrace,
+	}
+
+	vehicle.sniffer.Listen(dataChan)
+
+	return vehicle
+}
+
+func createSniffer(global excel_models.GlobalInfo, config Config, trace zerolog.Logger) sniffer.Sniffer {
```
jmaralo, line 71: put this and `getFilter` in the sniffer module

src/vehicle/constructor.go
```diff
@@ -0,0 +1,251 @@
...
+	sniffer, err := sniffer.New(filter, config.Network.Sniffer)
+
+	if err != nil {
+		trace.Fatal().Stack().Err(err).Msg("error creating sniffer")
+	}
+
+	return *sniffer
+}
+
+func createPipes(global excel_models.GlobalInfo, dataChan chan<- packet.Packet, onConnectionChange func(string, bool), config Config, trace zerolog.Logger) map[string]*pipe.Pipe {
```
jmaralo, line 82: put this and `getOnConnectionChange` in the pipe module

src/vehicle/constructor.go
```diff
@@ -0,0 +1,251 @@
...
+	return pipes
+}
+
+func getOnConnectionChange(board string, onConnectionChange func(string, bool)) func(bool) {
+	return func(state bool) {
+		onConnectionChange(board, state)
+	}
+}
+
+func createPacketParser(global excel_models.GlobalInfo, boards map[string]excel_models.Board, trace zerolog.Logger) (packet_parser.PacketParser, error) {
```
jmaralo, line 103: put this and all the necessary functions in the packet parser module

src/vehicle/constructor.go
```diff
@@ -0,0 +1,251 @@
...
+		tcpAddrDst = fmt.Sprintf("%s or (dst host %s)", tcpAddrDst, addr)
+	}
+	tcpAddrSrc = strings.TrimPrefix(tcpAddrSrc, " or ")
+	tcpAddrDst = strings.TrimPrefix(tcpAddrDst, " or ")
+	tcp = fmt.Sprintf("%s and (%s) and (%s)", tcp, tcpAddrSrc, tcpAddrDst)
+
+	filter := fmt.Sprintf("(%s) or (%s) or (%s)", ipip, udp, tcp)
+	trace.Trace().Strs("addrs", addrs).Str("filter", filter).Msg("new filter")
+	return filter
+}
```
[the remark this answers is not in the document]
jmaralo: split into three functions: `getIPIPfilter`, `getUDPFilter` and `getTCPFilter`

src/vehicle/vehicle.go
```diff
@@ -1,126 +1,172 @@
...
 	trace zerolog.Logger
 }
 
-func (vehicle *Vehicle) Listen(updateChan chan<- models.Update, messagesChan chan<- interface{}) {
-	vehicle.trace.Info().Msg("start listening")
-	go func() {
-		for raw := range vehicle.updateChan {
-			rawCopy := make([]byte, len(raw))
-			copy(rawCopy, raw)
+func (vehicle *Vehicle) Listen(updateChan chan<- models.PacketUpdate, protectionChan chan<- models.Protection) {
```
jmaralo, line 38: split into two functions

src/vehicle/vehicle.go
```diff
@@ -1,126 +1,172 @@
...
+	vehicle.trace.Debug().Msg("vehicle listening")
+	for raw := range vehicle.dataChan {
+		payloadCopy := make([]byte, len(raw.Payload))
+		copy(payloadCopy, raw.Payload)
+
+		switch id := raw.Metadata.ID; {
```
jmaralo, line 44: This is not decoding orders but it could receive orders

src/vehicle/vehicle.go
```diff
@@ -1,126 +1,172 @@
...
+
 			if err != nil {
-				vehicle.trace.Error().Stack().Err(err).Str("raw", fmt.Sprintf("%#v", string(raw))).Msg("parse message")
+				vehicle.trace.Error().Err(err).Msg("error decoding protection")
 				continue
 			}
-			messagesChan <- msg
+			protectionChan <- protection
+		default:
+			fmt.Println("UNEXPECTED VALUE")
```
jmaralo, line 67: use trace instead

src/vehicle/vehicle.go
```diff
@@ -1,126 +1,172 @@
...
 		}
-	}()
 
+	}
 }
 
 func (vehicle *Vehicle) SendOrder(order models.Order) error {
```
jmaralo, line 73: missing bitarray encoding (even if FW isn't using it put the code and we can comment it if we don't want it)

src/vehicle/vehicle.go
```diff
@@ -1,126 +1,172 @@
...
 	return err
 }
 
-func fieldsValuesToMap(fields map[string]models.Field) map[string]any {
-	valuesMap := make(map[string]any)
+func (vehicle *Vehicle) applyUnitConversion(values map[string]packet.Value) map[string]packet.Value {
+	newValues := make(map[string]packet.Value)
+
+	for name, value := range values {
+		switch typedValue := value.(type) {
```
jmaralo, line 107: put the switch in its own function

src/vehicle/vehicle.go
```diff
@@ -1,126 +1,172 @@
...
+	for name, field := range order.Fields {
+		switch value := field.Value.(type) {
+		case float64:
+			values[name] = packet.Numeric(value)
+		case bool:
+			values[name] = packet.Boolean(value)
+		case string:
+			values[name] = packet.Enum(value)
+		default:
+			trace.Error().Msg("order field value not recognized")
```
jmaralo, line 157: store the type (with `.Str("type", fmt.Sprintf("%T", field.Value))`) and the name (with `.Str("name", name)`)

jmaralo (changes requested):
It's ok, but I prefer the previous implementation of the parser where each ID had a kind associated and then a map from kind to decoder and from kind to encoder, it was more flexible and easier to use. Then to use it with the rest you just had to convert the internal parser types to PacketUpdate, Order, etc.

I feel most changes to vehicle just make it harder to read and make functions longer.

If you don't want to roll back some changes is alright because this will work and not much more changes need to be made, but that is my general impression

lipezaballa (approved):

PR #100, opened 2023-04-26
[PR 1] DEV-157 interface input selection

jmaralo:
The new interface prompts the user for the interface they want to use for the sniffer. It is a list of all the detected interfaces with their name, description and addresses. Then the users enters the index of the list they want to use.

Once the interface is selected, the execution continues as before

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -1,7 +1,9 @@
 package main
 
 import (
+	"bufio"
 	"flag"
+	"fmt"
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


src/main.go
```diff
@@ -184,3 +191,42 @@ func getConfig(path string) Config {
...
 }
+
+func selectDev() (pcap.Interface, error) {
+	devs, err := pcap.FindAllDevs()
+	if err != nil {
+		return pcap.Interface{}, err
+	}
+
+	fmt.Printf("select a device: (0-%d)\n", len(devs)-1)
+	for i, dev := range devs {
```
smorsua, line 202: Put loop in separate function

smorsua (commented):
lgtm, i'll approve when you make the changes

lipezaballa (approved):

PR #101, opened 2023-04-26
[PR1] DEV-158 New Protection model

smorsua:
This PR changes the ProtectionMessage model. It is now a JSON and need to be parsed differently. This is and example of the structure of a ProtectionMessage:

```
{
    "boardId": 12,
    "timestamp": {
        "counter": 38923221,
        "second": 32,
        "minute": 12,
        "hour": 8,
        "day": 25,
        "month": 7,
        "year": 2023
    },
    "protection": {
        "name": "VCELL1",
        "type": 4412323,
        "data": {
            "value": 3.3,
            "bounds": [10, 20]
        }
    }
}
```


src/protection_logger/loggable_protection.go
```diff
@@ -6,36 +6,36 @@ import (
...
-		violationData = fmt.Sprintf("%s Got: %f Want: %f", violation.Kind, violation.Got, violation.Want)
-	case vehicle_models.LowerBoundViolation:
-		violationData = fmt.Sprintf("%s Got: %f Want: > %f", violation.Kind, violation.Got, violation.Want)
-	case vehicle_models.UpperBoundViolation:
-		violationData = fmt.Sprintf("%s Got: %f Want: < %f", violation.Kind, violation.Got, violation.Want)
-	case vehicle_models.EqualsViolation:
-		violationData = fmt.Sprintf("%s %f is not allowed", violation.Kind, violation.Got)
-	case vehicle_models.NotEqualsViolation:
-		violationData = fmt.Sprintf("%s %f should be %f", violation.Kind, violation.Got, violation.Want)
+func getDataString(data any) string {
```
jmaralo, line 23:
```suggestion
func getDataString(data any) string {
    switch data := data.(type) {
	case vehicle_models.OutOfBounds:
		return fmt.Sprintf("Got: %f Want: %f", data.Value, data.Bounds)
	case vehicle_models.LowerBound:
		return fmt.Sprintf("Got: %f Want: > %f", data.Value, data.Bound)
	case vehicle_models.UpperBound:
		return fmt.Sprintf("Got: %f Want: < %f", data.Value, data.Bound)
	case vehicle_models.Equals:
		return fmt.Sprintf("%f is not allowed", data.Value)
	case vehicle_models.NotEquals:
		return fmt.Sprintf("%f should be %f", data.Value, data.Want)
	default:
		violationData = fmt.Sprintf("%s UNRECOGNIZED VIOLATION", violation.Type())
		return fmt.Sprintf("UNRECOGNIZED VIOLATION: %v", data)
	}
}
```

src/vehicle/protection_parser/protection_parser.go
```diff
@@ -0,0 +1,99 @@
...
+		},
+	}, nil
+}
+
+func (parser *ProtectionParser) getKind(id uint16) (string, error) {
+	if id == parser.faultId {
+		return "fault", nil
+	} else if id == parser.warningId {
+		return "warning", nil
+	} else {
```
jmaralo, line 78:
```suggestion
	}
```

jmaralo (dismissed):
Just two little things to change, but its great

lipezaballa (approved):

src/vehicle/protection_parser/protection_parser.go
Losina24: also u can replace all else if for if directly

Losina24 (approved):

PR #102, opened 2023-05-08
[PR 1] DEV-161 Ade-linter, PID checker and changes to loggables.

smorsua:
This PR integrates the ade-linter, a package that finds errors in ADE. It also adds the PID checker, a tiny package that makes sure only one instance of a program is running. Finally, some changes to the loggables have been made. This means that we've changed what is written into the log files. The pid is a rework of the [soellman/pidfile](https://pkg.go.dev/github.com/soellman/pidfile) package.


src/excel_adapter/excel_adapter.go
```diff
@@ -68,9 +72,31 @@ func fetchDocument(downloadConfig internals.DownloadConfig, parseConfig internal
...
+		var selection string
+		fmt.Scanln(&selection)
+
+		if strings.ToUpper(selection) == "Y" || strings.ToUpper(selection) == "YES" {
+			return true
+		} else if strings.ToUpper(selection) == "N" || strings.ToUpper(selection) == "NO" {
+			return false
+		} else {
+			fmt.Println("Incorrect input")
+		}
```
jmaralo, line 96: add a hint that you can input something again, like `> ` or some thing like that

src/excel_adapter/internals/excel_retriever.go
```diff
@@ -9,20 +9,24 @@ import (
...
-	Path        string
-	Name        string
-	Credentials string
+	Id   string
+	Path string
+	Name string
 }
 
+//go:embed secret.json
+var api_key []byte
```
jmaralo, line 25: don't use snake case for variables
```suggestion
var apiKey []byte
```

src/excel_adapter/internals/excel_retriever.go
```diff
@@ -9,20 +9,24 @@ import (
...
+
 func DownloadFile(config DownloadConfig) error {
 	trace.Trace().Str("id", config.Id).Str("path", config.Path).Str("name", config.Name).Msg("download file")
-	client, errClient := getClient(config.Credentials)
+	client, errClient := getClient(api_key)
```
jmaralo, line 29:
```suggestion
	client, errClient := getClient(apiKey)
```

src/logger_handler/logger_handler.go
```diff
@@ -54,42 +53,35 @@ func (handler *LoggerHandler) Log(loggable Loggable) {
...
 	case handler.config.Topics.Enable:
-		handler.handleEnable(payload, source)
-	}
-
-	handler.notifyState()
-}
+		var enable bool
+		err := json.Unmarshal(payload, &enable)
+		if err != nil {
+			handler.trace.Error().Stack().Err(err).Msg("unmarshal enable")
```
jmaralo, line 62: add a return because if not a random message can change the logger state
```suggestion
			handler.trace.Error().Stack().Err(err).Msg("unmarshal enable")
			return
```

src/logger_handler/logger_handler.go
```diff
@@ -54,42 +53,35 @@ func (handler *LoggerHandler) Log(loggable Loggable) {
...
 	}
 
 	handler.isRunningMx.Lock()
 	if enable && !handler.isRunning {
+		handler.isRunningMx.Unlock()
 		handler.start()
 	} else if !enable && handler.isRunning {
+		handler.isRunningMx.Unlock()
 		handler.stop()
 	}
```
jmaralo, line 84: put this in a separate function to defer the mutex unlock, it's more clear and less error prone
```suggestion
func (handler *LoggerHandler) changeState(enable bool) {
	handler.isRunningMx.Lock()
	defer handler.isRunningMx.Unlock()
	if enable && !handler.isRunning {
		handler.start()
	} else if !enable && handler.isRunning {
		handler.stop()
	}
}
```

src/main.go
```diff
@@ -140,59 +149,19 @@ func main() {
...
-func startOrderRoutine(orderChannel <-chan vehicle_models.Order, vehicle *vehicle.Vehicle, loggerHandler *logger_handler.LoggerHandler) {
-	for ord := range orderChannel {
-		err := vehicle.SendOrder(ord)
-
-		if err != nil {
-			trace.Error().Any("order", ord).Msg("error sending order")
+	if err != nil {
+		switch err {
+		case ErrProcessRunning:
+			trace.Fatal().Err(err).Msg("BACKEND IS ALREADY RUNNING")
```
jmaralo, line 158: There is no need to capitalize the message
```suggestion
			trace.Fatal().Err(err).Msg("Backend is already running")
```

src/order_logger/loggable_order.go
```diff
@@ -6,18 +6,23 @@ import (
...
+type LoggableOrder models.Order
 
 func (lo LoggableOrder) Id() string {
-	return fmt.Sprint(lo.Metadata.ID)
+	return fmt.Sprint(lo.ID)
 }
 
 func (lo LoggableOrder) Log() []string {
-	log := []string{fmt.Sprint(lo.Metadata.ID)}
+	return []string{fmt.Sprintf("[FROM GUI]: %d %v", lo.ID, lo.Fields)}
```
jmaralo, line 16: I don't like this format for the log file, since we use CSV we should try to log it like that. This just creates one column which is difficult to process later, instead, try to split the information into columns. For our data what I thing is best is to have the following columns: timestamp, from, to, seqNum, id and values. In this case you don't have all the information, that is fine, just fill the timestamp (with `time.Now()`), from (set it to `control` or something like that), id and values and leave the others empty.

src/order_logger/loggable_order.go
```diff
@@ -6,18 +6,23 @@ import (
...
+	return fmt.Sprint(lto.Metadata.ID)
+}
 
-	for name, value := range lo.Values {
-		log = append(log, name, fmt.Sprint(value))
-	}
+func (lto LoggableTransmittedOrder) Log() []string {
 
-	return log
+	return []string{fmt.Sprintf("[TRANSMITTED]: %v %s %s %d %d %v", lto.Metadata.Timestamp, lto.Metadata.From, lto.Metadata.To, lto.Metadata.SeqNum, lto.Metadata.ID, lto.Values)}
```
jmaralo, line 27: same as before

src/pid.go
```diff
@@ -0,0 +1,83 @@
...
+
+	err = process.Signal(syscall.Signal(0))
+
+	if err != nil && err.Error() == "no such process" {
+		return false
+	}
+
+	if err != nil && err.Error() == "os: process already finished" {
+		return false
+	}
```
jmaralo, line 75: there should be a better way to tell apart the errors, iirc depending on the os language they can be in other languages.

src/vehicle/constructor.go
```diff
@@ -125,3 +139,39 @@ func getIdToBoard(boards map[string]excel_models.Board, trace zerolog.Logger) ma
...
+func getBoardIdsFromType(boards map[string]excel_models.Board, kind string) common.Set[uint16] {
+	ids := common.NewSet[uint16]()
+
+	for _, board := range boards {
+		for _, packet := range board.Packets {
+			if packet.Description.Type == kind {
+				id, err := strconv.ParseInt(packet.Description.ID, 10, 16)
+
+				if err != nil {
+					//TODO:
```
jmaralo, line 152: solve TODO

src/vehicle/models/sorted_packets.go
```diff
@@ -0,0 +1,17 @@
...
+}
+
+func (packets SortedPackets) Less(i, j int) bool {
+	return int(packets[i].ID) < int(packets[j].ID)
+}
+
+func (packets SortedPackets) Swap(i, j int) {
+	copyPacket := packets[i]
+	packets[i] = packets[j]
+	packets[j] = copyPacket
```
jmaralo, line 16:
```suggestion
	packets[i], packets[j] = packets[j], packets[i]
```

src/vehicle/models/time_accumulation.go
```diff
@@ -0,0 +1,6 @@
+package models
+
+type TimeAccumulation struct {
+	Value     float64 `json:"value"`
+	TimeLimit float64 `json:"timelimit"`
```
jmaralo, line 5: this is missing `bound`, I've sent you to slack the structure Stefan sent me

jmaralo (changes requested):
Great job, just some things to improve the code.

Also lets try to keep the PRs smaller (not for this specific PR, but because we have been doing huge PRs recently).

jmaralo (approved):
lgtm, I've made a little comment that doesn't need to be changed, but it is the usual way to do things

src/pid.go
```diff
@@ -0,0 +1,83 @@
...
+	"syscall"
+)
+
+var (
+	ErrProcessRunning = errors.New("process is running")
+)
+
+// Write writes a pidfile, returning an error
+// if the process is already running or pidfile is orphaned
+func WritePid(filename string) error {
```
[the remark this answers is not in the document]
smorsua: This code came from a package so the comment is from there, i'll remove it

src/main.go
```diff
@@ -259,3 +228,58 @@ func acceptInput(limit int) (int, error) {
...
+	if fileErr != nil {
+		trace.Fatal().Stack().Err(fileErr).Msg("error reading config file")
+	}
+
+	reader := strings.NewReader(string(configFile))
+
+	var config Config
+
+	// decodeErr := toml.NewDecoder(reader).DisallowUnknownFields().Decode(&config)
+	decodeErr := toml.NewDecoder(reader).Decode(&config)
```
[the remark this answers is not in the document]
smorsua: I'll change it to a todo

src/pid.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return nil
+}
+
+func getOldPid(filename string) (int, error) {
+	oldPidFile, err := os.Open(filename)
+	if err != nil {
+
+		return 0, err
+	}
```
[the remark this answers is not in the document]
smorsua: Given that each function can fail for a different reason, I prefer to leave it like this in this case

lipezaballa (approved):

PR #107, opened 2023-05-14
[PR 1] DEV-170 limit the max number of websocket connections

jmaralo:
This PR updates a little bit the websocket broker logic to limit the max users that can connect at the same time, it uses the length of the connection map, that gets updated each time a client connects/disconnects.


lipezaballa (approved):

jmaralo:
Closing because it gets outdated by DEV-171, which changes this logic

PR #108, opened 2023-05-15
[PR 1] DEV-171 add support for multiple HTTP servers and rate limit

jmaralo:
This PR adds the ability to add or remove http servers via `config.toml`. Each server has its own endpoints, static path and client limit (optional). To configure a server add a section to the `config.toml` like this:

```toml
[server.audience]
address = "172.16.0.2:4000"
client_limit = 2
static = "./static/audience"

[server.audience.endpoints]
pod_data = "/podDataStructure"
order_data = "/orderStructures"
programable_boards = "/uploadableBoards"
connections = "/backend"
file_server = "/"
```

In the example the new server will be called audience, it will be hosted on 172.16.0.2:4000 and will have a client limit of 2. If the client limit is omited, it has no limit to the number of clients that can connect.

`websocket_broker` has been changed to accommodate for the changes, it now notifies through a channel when a client has been disconnected so the number of connected clients can be tracked correctly.

### DIFF
--- a/src/config.go
+++ b/src/config.go
@@ -29,6 +29,4 @@ type Config struct {
 	Messages message_transfer.MessageTransferConfig
 	Server   server.Config
 	BLCU     blcu.BLCUConfig `toml:"blcu"`
-
-	Test map[string]uint `toml:"test"`
 }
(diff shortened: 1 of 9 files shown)


smorsua (approved):
LGTM, lets investigate if the net and websocket package offer something for keep alives instead of implementing it ourselves.

lipezaballa (approved):

PR #111, opened 2023-05-23
[PR 1] DEV-177 BLCU fixes

jmaralo:
After integrating the BLCU logic with FW, here are the fixes necessary to make it work, it also features an update to notify the progress to the frontend. However, This is currently not supported by the frontend and will cause issues.

### DIFF
--- a/src/blcu/download.go
+++ b/src/blcu/download.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"time"
@@ -17,34 +18,23 @@ type downloadRequest struct {
 	Board string `json:"board"`
 }
 
-func (blcu *BLCU) download(payload json.RawMessage) (string, DownloadResult) {
+func (blcu *BLCU) download(payload json.RawMessage) (string, []byte, error) {
 	blcu.trace.Debug().Msg("Handling download")
 	var request downloadRequest
 	if err := json.Unmarshal(payload, &request); err != nil {
 		blcu.trace.Error().Err(err).Stack().Msg("Unmarshal payload")
-		return "", DownloadResult{Err: err}
+		return "", nil, err
 	}
 
 	if err := blcu.requestDownload(request.Board); err != nil {
 		blcu.trace.Error().Err(err).Stack().Msg("Request download")
-		return request.Board, DownloadResult{Err: err}
+		return request.Board, nil, err
 	}
 
-	download, err := blcu.ReadTFTP()
-	if err != nil {
-		blcu.trace.Error().Err(err).Stack().Msg("Read payload")
-		return request.Board, DownloadResult{Err: err}
-	}
+	buffer := &bytes.Buffer{}
+	err := blcu.ReadTFTP(buffer, blcu.notifyDownloadProgress)
 
-	go blcu.consumeDownloadPercentage(download.Percentage)
-
-	return request.Board, <-download.Result
-}
-
-func (blcu *BLCU) consumeDownloadPercentage(percentage <-chan float64) {
-	for p := range percentage {
-		blcu.notifyDownloadProgress(p)
-	}
+	return request.Board, buffer.Bytes(), err
 }
 
 func (blcu *BLCU) requestDownload(board string) error {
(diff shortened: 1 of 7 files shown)


src/blcu/download.go
```diff
@@ -75,43 +75,50 @@ func (blcu *BLCU) createDownloadOrder(board string) models.Order {
...
 	if err != nil {
-		return err
+		return Download{}, err
 	}
 
-	_, err = receiver.WriteTo(writer)
-	if err != nil {
-		return err
-	}
+	download := NewDownload(786432) // Size of the flash memory (downlaod size)
```
smorsua, line 91: Put size in const FlashMemorySize

src/blcu/download.go
```diff
@@ -17,34 +17,34 @@ type downloadRequest struct {
...
 		blcu.trace.Error().Err(err).Stack().Msg("Request download")
-		return downloadData{}, err
+		return request.Board, DownloadResult{Err: err}
 	}
 
-	buffer := bytes.NewBuffer([]byte{})
-	if err := blcu.ReadTFTP(buffer); err != nil {
-		blcu.trace.Error().Err(err).Stack().Msg("Read TFTP")
-		return downloadData{}, err
+	download, err := blcu.ReadTFTP()
```
smorsua, line 33: I would do the following: ReadTFTP blocks until it downloads or errors and it accepts a callback (onProgress), where you receive the progress and can whatever you want to do with it.

src/blcu/download.go
```diff
@@ -75,43 +75,50 @@ func (blcu *BLCU) createDownloadOrder(board string) models.Order {
...
 
-	return nil
+	return download, nil
 }
 
 type downloadResponse struct {
-	IsSuccess bool   `json:"success"`
-	File      []byte `json:"file,omitempty"`
+	Percentage float64 `json:"percentage"`
+	IsSuccess  *bool   `json:"success,omitempty"`
```
smorsua, line 99: Make IsSuccess value and send false while percentage is less than 100

src/blcu/download.go
```diff
@@ -123,5 +130,80 @@ func (blcu *BLCU) writeDownloadFile(data downloadData) error {
...
-	return os.WriteFile(path.Join(blcu.config.DownloadPath, data.Board+".bin"), data.Payload, 0777)
+	return os.WriteFile(path.Join(blcu.config.DownloadPath, fmt.Sprintf("%s-%d.bin", board, time.Now().Unix())), data, 0777)
+}
+
+type DownloadResult struct {
+	Data []byte
+	Err  error
+}
+
+type Download struct {
```
smorsua, line 141: Download would store the onProgress callback passed to ReadTFTP

src/blcu/download.go
```diff
@@ -123,5 +130,80 @@ func (blcu *BLCU) writeDownloadFile(data downloadData) error {
...
+func (download *Download) Write(p []byte) (n int, err error) {
+	if download.writeErr != nil {
+		return 0, download.writeErr
+	}
+	n, err = download.output.Write(p)
+	if err != nil {
+		download.writeErr = err
+		download.abort(err)
+	} else {
+		download.notify <- n
```
smorsua, line 181: You'd call onProgress here, with the percentage

src/blcu/download.go
```diff
@@ -123,5 +130,80 @@ func (blcu *BLCU) writeDownloadFile(data downloadData) error {
...
+	}
+	return n, err
+}
+
+func (download *Download) abort(err error) {
```
smorsua, line 186: Why do you need to send the error through a channel? receiver.WriteTo returns an error, so consumers should check that error to determine if the download failed

src/blcu/download.go
```diff
@@ -123,5 +130,80 @@ func (blcu *BLCU) writeDownloadFile(data downloadData) error {
...
+	close(download.notify)
+	download.resultChan <- DownloadResult{
```
smorsua, line 188: If receiver.WriteTo block, you dont need this channel

smorsua (changes requested):
Pretty clear implementation, I'd avoid concurrency in some places to avoid using many channels and such

smorsua (approved):
LGTM

lipezaballa (approved):

PR #113, opened 2023-05-27
[PR 1] DEV-181 Added observables

smorsua:
This PR adds observables to the different modules that clients can subscribe to. The result it's the same as before, broadcasting messages, but it's more declarative and the same logic is shared across all places that offer a subscription.
- The `ReplayObservable` sends the last stored value when an observer subscribes and broadcast future values.
- The `NoReplayObservable` doesn't send the last stored value (in fact, it doesn't store anything) and broadcasts future values.
- The `WsObserver` satisfies the Observer interface, and sends a message thorugh websocket when an update is received.


jmaralo (approved):
LGTM, there are some modules that don't use the config after the changes, but I would implement it in a later PR once this and DEV-178 are merged to develop

lipezaballa (approved):

PR #115, opened 2023-06-02
[PR 1] Fixed ws_handler deadlock, pipe_readers and simple subscription

smorsua:
The pipe_readers are structs that read from the pipes in a particular way. Each pipe has a map that goes from id (first two bytes of any message) to reader. Right now, we have two types of readers:
- For JSON messages from the boards, we read until we find a 0x00.
- For StateOrder messages, part of the message is the number of orders it carries, and so we can calculate the size of the message i.e the amount of bytes we have to read.

A deadlock has been fixed in the ws_handler. The code to subscribe to a module is now common to all modules with a subscription.

Finally, the subcriptions from the frontend now provide an id for each subcription to avoid errors relating to the same connection unsubscribing and subscribing to the same module in shorts periods of time (if the order was not correct, the subscription could have failed).

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -248,7 +248,7 @@ func acceptInput(limit int) (int, error) {
 		}
 
 		if dev < 0 || dev >= limit {
-			red.Println("invalid device selected\n")
+			red.Println("invalid device selected")
 			continue
 		} else {
 			return dev, nil
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -1,0 +1,49 @@
+name: Create Release
+
+on:
+  push:
+    branches:
+      - master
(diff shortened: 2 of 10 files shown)


jmaralo (changes requested):
Just two minor things

src/main.go
```diff
@@ -133,7 +134,12 @@ func main() {
 
 	go func() {
 		for stateOrders := range stateOrdersChan {
-			orderTransfer.UpdateStateOrders(stateOrders)
+			switch stateOrders.Action {
+			case "addStateOrders":
```
smorsua: Put values in const

smorsua (commented):
lgtm

lipezaballa (approved):

